PR #15, opened 2021-02-23
Np 2157 use commons

torarnet:


src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
```diff
@@ -45,75 +36,70 @@
...
         this.cristinApiClient = cristinApiClient;
     }
 
-    public void setCristinApiClient(CristinApiClient cristinApiClient) {
-        this.cristinApiClient = cristinApiClient;
+    @Override
+    protected ProjectPresentation[] processInput(Void input, RequestInfo requestInfo, Context context)
+        throws ApiGatewayException {
+
+        return createProjectPresentations(checkLanguageAndFetch(requestInfo), checkTitleAndFetch(requestInfo));
```
axthosarouris, line 59: assign the parameters to variables for better readbility.

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
```diff
@@ -45,75 +36,70 @@
...
+        List<Project> projects = attempt(() ->
+            cristinApiClient.queryAndEnrichProjects(cristinQueryParameters, language)).orElseThrow();
 
-        String language = queryStringParameters.getOrDefault(LANGUAGE_KEY, DEFAULT_LANGUAGE_CODE);
-        if (!VALID_LANGUAGE_CODES.contains(language)) {
-            throw new RuntimeException(LANGUAGE_INVALID);
-        }
+        return projects.stream()
+            .map(project -> presentationConverter.asProjectPresentation(project, language))
+            .collect(Collectors.toList());
```
axthosarouris, line 102: refactor to method `transfromProjectsToProjectPresentations` ?

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
```diff
@@ -128,10 +114,14 @@ private boolean isValidTitle(String str) {
 
     private boolean isValidCharacter(char c) {
         return Character.isWhitespace(c)
-                || Character.isLetterOrDigit(c)
-                || c == CHARACTER_DASH
-                || c == CHARACTER_COMMA
-                || c == CHARACTER_PERIOD;
+            || Character.isLetterOrDigit(c)
+            || c == CHARACTER_DASH
```
axthosarouris, line 118: These 3 chars could be a set of invalid characters and then you only  need to check INVALID_CHARS.contains(c)

axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
torarnet: Those 3 chars are actually valid characters in addition to letter, digit and whitespace. But i can make a set of those character and do a VALID_CHARS.contains(c)

build.gradle
```diff
@@ -13,11 +13,30 @@ targetCompatibility = JavaVersion.VERSION_11 // bytecode target version
 
 repositories {
     mavenCentral()
+    maven { url  "https://dl.bintray.com/unit-nva/nva-commons" }
+    maven { url 'https://jitpack.io' }
+}
+
+project.ext{
+    nvaCommonsVersion = '1.0.4'
```
brinxmat, line 21: It surprises me that jupiterVersion is not in project.ext, but this is not to say that it is incorrect

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
```diff
@@ -45,75 +36,70 @@
...
+    private String checkLanguageAndFetch(RequestInfo requestInfo) throws BadRequestException {
+        return Optional.of(getQueryParamOpt(requestInfo, LANGUAGE_QUERY_PARAMETER)
+            .orElse(DEFAULT_LANGUAGE_CODE))
+            .filter(this::isValidLanguage)
+            .orElseThrow(() -> new BadRequestException(LANGUAGE_INVALID));
+    }
 
-            Type projectListType = new TypeToken<ArrayList<ProjectPresentation>>() {
-            }.getType();
+    protected Optional<String> getQueryParamOpt(RequestInfo requestInfo, String queryParameter) {
```
brinxmat, line 80: getQueryParameter, the return type is clear from the method definition

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
```diff
@@ -45,75 +36,70 @@
...
-        Map<String, String> queryStringParameters = (Map<String, String>) input.get(QUERY_STRING_PARAMETERS_KEY);
-        String title = queryStringParameters.get(TITLE_KEY);
-        String language = queryStringParameters.getOrDefault(LANGUAGE_KEY, DEFAULT_LANGUAGE_CODE);
+    protected Integer getSuccessStatusCode(Void input, ProjectPresentation[] output) {
+        return HttpURLConnection.HTTP_OK;
+    }
 
-        try {
-            Map<String, String> cristinQueryParameters = createCristinQueryParameters(title, language);
+    private String checkTitleAndFetch(RequestInfo requestInfo) throws BadRequestException {
```
brinxmat, line 67: getValidTitle (I don't like methods that tell me they do multiple things with "and" as this tends to mean the method is violating the SRP), I also tend to reserve "fetch" for HTTP-GETs.

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
```diff
@@ -45,75 +36,70 @@
...
-            gatewayResponse.setStatusCode(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode());
-            gatewayResponse.setErrorBody(e.getMessage());
-        }
+    private ProjectPresentation[] createProjectPresentations(String language, String title) {
+        List<ProjectPresentation> projectPresentations = createProjectPresentationList(language, title);
+        return convertToArray(projectPresentations);
+    }
 
-        return gatewayResponse;
+    private ProjectPresentation[] convertToArray(List<ProjectPresentation> projectPresentations) {
```
brinxmat, line 89: Probably just
        return projectPresentations.toArray(ProjectPresentation[]::new);

brinxmat (changes requested):
In general, very tidy, well formulated code. Follows our standards

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
torarnet: Seems char array works a little different, so after discussing it with Orestis we decided to leave it as it is

build.gradle
```diff
@@ -13,11 +13,30 @@ targetCompatibility = JavaVersion.VERSION_11 // bytecode target version
...
+    jacksonVersion = '2.12.0'
 }
 
 def jupiterVersion = "5.6.0"
 
 dependencies {
+    testImplementation 'com.github.BIBSYSDEV:nva-testutils:0.1.18'
```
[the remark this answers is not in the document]
torarnet: sort them last in each group or put every testDependency last under all other dependencies?

build.gradle
torarnet: Should i put it there?

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
torarnet: should i only change name or split logic into one validator and one get value as well? I wanted to do several things with less code, and calling this::isValid implicit made the code a little more compact

build.gradle
brinxmat: Why not? :)

brinxmat (approved):

build.gradle
torarnet: Moved it to project.ext

axthosarouris (approved):

PR #17, opened 2021-02-25
NP-2157: Initial change from Gson to Jackson

torarnet:
Just first step, will refactor model classes more in later work
Extracted method for enriching projects. Still complains about using unneeded local variable

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,8 +19,8 @@ repositories {
 
 project.ext{
     nvaCommonsVersion = '1.1.1'
-    jacksonVersion = '2.12.0'
-    jupiterVersion = '5.6.0'
+    jacksonVersion = '2.12.1'
+    jupiterVersion = '5.7.1'
 }
 
 dependencies {
@@ -45,6 +45,8 @@ dependencies {
 
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '3.2.4'
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: project.ext.jupiterVersion
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params', version: project.ext.jupiterVersion
+
     testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
     testImplementation 'com.github.BIBSYSDEV:nva-testutils:0.1.18'
     testImplementation group: 'no.unit.nva', name: 'logutils', version: project.ext.nvaCommonsVersion
@@ -100,7 +102,7 @@ jacocoTestCoverageVerification {
             limit {
                 counter = 'METHOD'
                 value = 'COVEREDRATIO'
-                minimum = 0.800
+                minimum = 0.87
             }
         }
 
@@ -108,7 +110,7 @@ jacocoTestCoverageVerification {
             limit {
                 counter = 'CLASS'
                 value = 'COVEREDRATIO'
-                minimum = 0.800
+                minimum = 0.86
             }
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,6 @@ dependencies {
 
     implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.9'
     implementation group: 'javax.ws.rs', name: 'javax.ws.rs-api', version: '2.1.1'
-    implementation group: 'com.google.code.gson', name: 'gson', version: '2.8.6'
     implementation group: 'org.slf4j', name: 'slf4j-api', version: '1.7.29'
     implementation group: 'org.jsoup', name: 'jsoup', version: '1.12.1'
     implementation 'com.amazonaws:aws-lambda-java-core:1.2.0'
--- a/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
+++ b/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
@@ -12,20 +12,18 @@ import java.net.URL;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
-import nva.commons.core.Environment;
 import nva.commons.core.JsonUtils;
 import org.apache.http.client.utils.URIBuilder;
 
 public class CristinApiClient {
 
     private static final String HTTPS = "https";
-    private static final String CRISTIN_API_HOST_ENV = "CRISTIN_API_HOST";
     private static final String CRISTIN_API_PROJECTS_PATH = "/v2/projects/";
     private static final ObjectMapper OBJECT_MAPPER = JsonUtils.objectMapper;
     private final transient String cristinApiHost;
 
-    public CristinApiClient(Environment environment) {
-        cristinApiHost = environment.readEnv(CRISTIN_API_HOST_ENV);
+    public CristinApiClient(String cristinApiHost) {
+        this.cristinApiHost = cristinApiHost;
     }
 
     protected static <T> T fromJson(InputStreamReader reader, Class<T> classOfT) throws IOException {
--- a/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
+++ b/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
@@ -1,6 +1,7 @@
 package no.unit.nva.cristin.projects;
 
 import static java.util.Arrays.asList;
+import static nva.commons.core.attempt.Try.attempt;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.io.InputStreamReader;
@@ -62,7 +63,7 @@ public class CristinApiClient {
     }
 
     protected URL generateQueryProjectsUrl(Map<String, String> parameters) throws MalformedURLException,
-            URISyntaxException {
+                                                                                  URISyntaxException {
         URIBuilder uri = new URIBuilder()
             .setScheme(HTTPS)
             .setHost(cristinApiHost)
@@ -84,15 +85,14 @@ public class CristinApiClient {
     }
 
     private List<Project> enrichProjects(String language, List<Project> projects) {
-        return projects.stream()
-            .map(project -> {
-                try {
-                    return getProject(project.cristinProjectId, language);
-                } catch (IOException | URISyntaxException e) {
-                    System.out.println("Error fetching cristin project with id: " + project.cristinProjectId);
-                }
+        return projects.stream().map(project -> enrichOneProject(language, project)).collect(Collectors.toList());
+    }
+
+    private Project enrichOneProject(String language, Project project) {
+        return attempt(() -> getProject(project.cristinProjectId, language))
+            .orElse((failure) -> {
+                System.out.println("Error fetching cristin project with id: " + project.cristinProjectId);
                 return project;
-            })
-            .collect(Collectors.toList());
+            });
     }
 }
--- a/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
+++ b/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
@@ -21,10 +21,10 @@ public class CristinApiClient {
     private static final String CRISTIN_API_HOST_ENV = "CRISTIN_API_HOST";
     private static final String CRISTIN_API_PROJECTS_PATH = "/v2/projects/";
     private static final ObjectMapper OBJECT_MAPPER = JsonUtils.objectMapper;
-    private final transient Environment environment;
+    private final transient String cristinApiHost;
 
     public CristinApiClient(Environment environment) {
-        this.environment = environment;
+        cristinApiHost = environment.readEnv(CRISTIN_API_HOST_ENV);
     }
 
     protected static <T> T fromJson(InputStreamReader reader, Class<T> classOfT) throws IOException {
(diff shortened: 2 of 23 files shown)


src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -75,21 +67,24 @@ protected URL generateQueryProjectsUrl(Map<String, String> parameters) throws Ma
 
     protected URL generateGetProjectUrl(String id, String language) throws MalformedURLException, URISyntaxException {
         URI uri = new URIBuilder()
-                .setScheme(HTTPS)
-                .setHost(CRISTIN_API_HOST)
-                .setPath(CRISTIN_API_PROJECTS_PATH + id)
-                .addParameter("lang", language)
-                .build();
+            .setScheme(HTTPS)
```
axthosarouris, line 70: not in scope and not very important but is there a possibility to build the URIs using the java constructors new URI()? The less libraries we use the faster the startup time. Of course if we need to use something, we should use it. Anyhow, I am just seeing it because of the linting, and it is outside the scope of this PR.

axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/ProjectPresentation.java
```diff
@@ -1,15 +1,19 @@
 package no.unit.nva.cristin.projects;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import java.util.ArrayList;
 import java.util.List;
 
+@JsonPropertyOrder({"cristin_project_id", "main_language"})
```
axthosarouris, line 8: You can make those constants since you are re-using them

axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
brinxmat, line 70: We will want to pass the host in as an AWS parameter because we may wish to change the target at some point between test/prod, or for any other number of reasons :D

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -75,21 +67,24 @@ protected URL generateQueryProjectsUrl(Map<String, String> parameters) throws Ma
...
     }
 
-    protected static <T> T fromJson(InputStreamReader reader, Class<T> classOfT) throws IOException {
-        try {
-            return new Gson().fromJson(reader, classOfT);
-        } catch (JsonSyntaxException e) {
-            final String s = e.getMessage() + " " + reader;
-            throw new IOException(s, e);
-        }
+    private List<Project> enrichProjects(String language, List<Project> projects) {
```
brinxmat, line 78: It's a bit weird that we consume a list of projects to produce a list of projects

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -75,21 +67,24 @@ protected URL generateQueryProjectsUrl(Map<String, String> parameters) throws Ma
...
+        return projects.stream()
+            .map(project -> {
```
brinxmat, line 80: In general, I react to the code in this class (it isn't code you have written) for a number of reasons:

 - it should take account for the fact that we want to download the projects from the Cristin API. rapidly, probably in parallel, but the code will lock for each request, accumulating time for each request
 - we have very little control over the requests because we are relying on URL.openStream
 - failures are ignored, which is unfortunate
 - we should have a chat with Orestis about how to avoid these issues practically in Lambdas (I have a code example, based on CompletableFuture, which I think will work, BUT may have side effects in lambdas)

brinxmat (changes requested):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
brinxmat: I realise some of my comments are not in scope — just looking at this code made me react in general, this isn't your fault @torarnet :D

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -14,29 +10,17 @@
...
-                        return getProject(project.cristinProjectId, language);
-                    } catch (IOException | URISyntaxException e) {
-                        System.out.println("Error fetching cristin project with id: " + project.cristinProjectId);
-                    }
-                    return project;
-                })
-                .collect(Collectors.toList());
-        return enrichedProjects;
+    protected static <T> T fromJson(InputStreamReader reader, Class<T> classOfT) throws IOException {
+        return JsonUtils.objectMapper.readValue(reader, classOfT);
```
brinxmat, line 23: Make the object mapper a constant

brinxmat (changes requested):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: Seems there is no easy way to do it and would also make code less readable as built in URI uses parameters like scheme and host in constructor only and there were no easy way to add query params, (it only takes in a string, we need to concatenate ourselves). We can do it but then we have to do a little more work with the query params, maybe in another pull request?

src/main/java/no/unit/nva/cristin/projects/ProjectPresentation.java
torarnet: Should i create constants for all strings used in annotations across all classes using them?

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: Should i set scheme and path as well in environment or only host?

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: it was there from before, i just extracted it into new method.

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: Maybe create a new Jira issue where we will improve it?

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
axthosarouris: Do not worry about it now

src/main/java/no/unit/nva/cristin/projects/ProjectPresentation.java
axthosarouris: Maybe they should be if they are used in many classes

axthosarouris (dismissed):

src/main/java/no/unit/nva/cristin/projects/ProjectPresentation.java
torarnet: I get this error when trying to put the names in variables: "Attribute value must be constant". Maybe you have to declare them directly as strings as is already done?

torarnet: wait, i figured it out, using static final it does not complain anymore. Will fix later before merging

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
```diff
@@ -38,12 +39,14 @@
...
     private static final String DEFAULT_LANGUAGE_CODE = "nb";
+    private static Environment environment = new Environment();
     private final transient CristinApiClient cristinApiClient;
     private final transient PresentationConverter presentationConverter = new PresentationConverter();
 
     @SuppressWarnings("unused")
+    @JacocoGenerated
     public FetchCristinProjects() {
-        this(new CristinApiClient(), new Environment());
+        this(new CristinApiClient(environment), environment);
```
janbj, line 49: Maybe read the environment variables here and pass them into the client constructor?

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -64,8 +64,8 @@ protected InputStreamReader fetchGetResult(URL url) throws IOException {
     protected URL generateQueryProjectsUrl(Map<String, String> parameters) throws MalformedURLException,
             URISyntaxException {
         URIBuilder uri = new URIBuilder()
-                .setScheme(HTTPS)
-                .setHost(CRISTIN_API_HOST)
+            .setScheme(HTTPS)
+            .setHost(environment.readEnv(CRISTIN_API_HOST_ENV))
```
janbj, line 68: Can we read the environment variable in the constructor instead to fail fast if config is wrong or missing?

janbj (changes requested):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: Should I pass the whole environment as parameter to constructor or only the environment variables i want to use?

axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
torarnet: Have to call `this` first in constructor otherwise compiler error. Therefore I have to send straight in to CristinApiClient

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: After discussion, only host for now

axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: Opened a issue on this

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -14,29 +11,24 @@
...
-                        return getProject(project.cristinProjectId, language);
-                    } catch (IOException | URISyntaxException e) {
-                        System.out.println("Error fetching cristin project with id: " + project.cristinProjectId);
-                    }
-                    return project;
-                })
-                .collect(Collectors.toList());
-        return enrichedProjects;
+    public CristinApiClient(Environment environment) {
+        cristinApiHost = environment.readEnv(CRISTIN_API_HOST_ENV);
```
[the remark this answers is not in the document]
brinxmat: In fact, the URI we send is in essence static except for the query parameters values for title and language and we use a lot of code to set these, so maybe just a String template with URI-encoded values. à la:

```"https://api.cristin.no/v2/projects/?title=%s&lang=%s&static=static&static=static"```

File: src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java (hunk 7 above)
Comment (janbj) [reply src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java]: I prefer the second one. I don't think the environment reader should leak into the client.

File: src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java (hunk 6 above)
Comment (janbj) [reply src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java:52]: Maybe you can create a helper method that creates the client?

File: src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java [hunk 9]
```diff
@@ -75,21 +75,24 @@ protected URL generateQueryProjectsUrl(Map<String, String> parameters) throws Ma
...
-        try {
-            return new Gson().fromJson(reader, classOfT);
-        } catch (JsonSyntaxException e) {
-            final String s = e.getMessage() + " " + reader;
-            throw new IOException(s, e);
-        }
+    private List<Project> enrichProjects(String language, List<Project> projects) {
+        return projects.stream()
+            .map(project -> {
+                try {
```
brinxmat, line 89: Use commons attempt

brinxmat (changes requested):

src/test/resources/api_query_response.json
```diff
@@ -0,0 +1,116 @@
+[ {
+  "cristin_project_id" : "2040291",
+  "main_language" : "en",
+  "titles" : [ {
+    "title" : "Opportunities and Challenges for Integrating Sámi Reindeer Herding Traditional Environmental Knowledge in Environmental Governance",
+    "language" : "en"
+  } ],
+  "participants" : [ {
+    "cristinPersonId" : "327970",
+    "fullName" : "Turi, Ellen Inga"
```
mattiahj, line 10: Is this a synthetic example? We should not be publishing personal information without the consent of the person in question (github is viewable by anyone).

build.gradle
```diff
@@ -101,15 +102,15 @@ jacocoTestCoverageVerification {
             limit {
                 counter = 'METHOD'
                 value = 'COVEREDRATIO'
-                minimum = 0.800
+                minimum = 0.87
```
mattiahj, line 105: For my sake, how are these ratios decided on and why?

brinxmat: In this case, it is just me upping the ratios as they improve — ideally 1.000, but as is, the code is only at 87%

src/main/java/no/unit/nva/cristin/projects/FetchCristinProjects.java
torarnet: Will do in another PR, have to get this one through, its already big enough

brinxmat (dismissed):

axthosarouris (dismissed):

janbj (dismissed):

janbj (approved):

brinxmat (approved):

PR #18, opened 2021-03-02
Np 2271 fix funding title

brinxmat:
Fixes the missing funder information, additionally:

- Adds some testing because the actual test for the data was missing
- bumped versions on Jackson and JUnit

### DIFF
--- a/src/main/java/no/unit/nva/cristin/projects/PresentationConverter.java
+++ b/src/main/java/no/unit/nva/cristin/projects/PresentationConverter.java
@@ -12,7 +12,7 @@ public class PresentationConverter {
         projectPresentation.cristinProjectId = project.cristinProjectId;
         projectPresentation.mainLanguage = project.mainLanguage;
 
-        Optional.ofNullable(project.title).orElse(new TreeMap<String, String>() {
+        Optional.ofNullable(project.title).orElse(new TreeMap<>() {
         }).forEach((key, value) -> {
             TitlePresentation titlePresentation = new TitlePresentation();
             titlePresentation.language = key;
@@ -41,7 +41,7 @@ public class PresentationConverter {
             FundingSourcePresentation fundingSourcePresentation = new FundingSourcePresentation();
             fundingSourcePresentation.fundingSourceCode = fundingSource.fundingSourceCode;
             fundingSourcePresentation.projectCode = fundingSource.projectCode;
-            fundingSource.fundingSourceName.entrySet().stream()
+            fundingSourcePresentation.names = fundingSource.fundingSourceName.entrySet().stream()
                     .map(name -> new FundingSourceNamePresentation(name.getKey(), name.getValue()))
                     .collect(Collectors.toList());
             projectPresentation.fundings.add(fundingSourcePresentation);
--- a/src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
+++ b/src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
@@ -59,7 +59,7 @@ public class FetchCristinProjectsTest {
     private FetchCristinProjects handler;
 
     @BeforeEach
-    public void setUp() {
+    void setUp() {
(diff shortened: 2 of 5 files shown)


src/main/java/no/unit/nva/cristin/projects/PresentationConverter.java
```diff
@@ -12,7 +12,7 @@ protected ProjectPresentation asProjectPresentation(Project project, String lang
         projectPresentation.cristinProjectId = project.cristinProjectId;
         projectPresentation.mainLanguage = project.mainLanguage;
 
-        Optional.ofNullable(project.title).orElse(new TreeMap<String, String>() {
+        Optional.ofNullable(project.title).orElse(new TreeMap<>() {
```
axthosarouris, line 15: Collections.emptyMap().
refactor to method

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -62,10 +67,19 @@ public void setUp() {
         handler = new FetchCristinProjects(mockCristinApiClient, environment);
     }
 
+    @ParameterizedTest
+    @ArgumentsSource(TestPairProvider.class)
+    void handlerReturnsExpectedBodyWhenRequestInputIsValid(String queryResponse,
```
axthosarouris, line 72: General comment: If this test ever breaks, it could be hard to find out why. But I see that there can be no quick solutions to this problem now.

src/main/java/no/unit/nva/cristin/projects/PresentationConverter.java
```diff
@@ -41,7 +41,7 @@ protected ProjectPresentation asProjectPresentation(Project project, String lang
             FundingSourcePresentation fundingSourcePresentation = new FundingSourcePresentation();
             fundingSourcePresentation.fundingSourceCode = fundingSource.fundingSourceCode;
             fundingSourcePresentation.projectCode = fundingSource.projectCode;
-            fundingSource.fundingSourceName.entrySet().stream()
+            fundingSourcePresentation.names = fundingSource.fundingSourceName.entrySet().stream()
```
axthosarouris, line 44: refactor ot method

brinxmat: Will refactor entirely in next PR

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
brinxmat: Yes. The test needs to be in place for next PR

src/main/java/no/unit/nva/cristin/projects/PresentationConverter.java
brinxmat: 1, actually not.
2. next PR

axthosarouris (approved):

src/test/java/no/unit/nva/cristin/projects/TestPairProvider.java
```diff
@@ -0,0 +1,29 @@
...
+            IoUtils.stringFromResources(Path.of("api_query_response.json"));
+
+    @Override
+    public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+        return Stream.of(
+                Arguments.of(CRISTIN_QUERY_PROJECTS_RESPONSE, CRISTIN_GET_PROJECT_RESPONSE, API_QUERY_RESPONSE_JSON)
+        );
+    }
+
+    private static InputStreamReader resourceAsInputStreamReader(String resource) {
```
torarnet, line 26: this is not in use?

brinxmat: Very true, it belongs in the next PR

axthosarouris (approved):

torarnet (approved):

PR #27, opened 2021-04-06
Np 2384 lookup one project

torarnet:


build.gradle
```diff
@@ -18,15 +18,15 @@ repositories {
 }
 
 project.ext{
-    nvaCommonsVersion = '1.1.1'
+    nvaCommonsVersion = '1.5.4'
```
axthosarouris, line 21: 1.5.10

src/main/java/no/unit/nva/cristin/projects/CommonUtils.java
```diff
@@ -0,0 +1,30 @@
...
+     * Validate content of a CristinProject.
+     *
+     * @param cristinProject the CristinProject to validate
+     * @return true if valid or false if not
+     */
+    public static boolean hasValidContent(CristinProject cristinProject) {
+        return cristinProject != null
+            && hasData(cristinProject.cristinProjectId)
+            && hasData(cristinProject.title);
+    }
```
axthosarouris, line 20: this belongs to the the CristinProject class.

src/main/java/no/unit/nva/cristin/projects/CommonUtils.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public static boolean hasData(String... strings) {
+        return Arrays.stream(strings).noneMatch(StringUtils::isBlank);
+    }
+
+    @SuppressWarnings("rawtypes")
+    public static boolean hasData(Map... maps) {
+        return Arrays.stream(maps).noneMatch(map -> map == null || map.isEmpty());
```
axthosarouris, line 28: same as above

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -66,12 +67,44 @@ public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetada
...
+            return new EmptyNvaProject();
+        }
+
+        NvaProject nvaProject = new NvaProjectBuilder(cristinProject).build();
+        nvaProject.setContext(Constants.PROJECT_LOOKUP_CONTEXT_URL);
+
+        return nvaProject;
+    }
+
+    private CristinProject attemptToGetCristinProject(String id, String language) {
```
axthosarouris, line 99: suggestion:
```
    private CristinProject attemptToGetCristinProject(String id, String language) {
        return attempt(() -> getProject(id, language))
                   .toOptional(failure-> logError(id, failure))
                   .orElse(null);
    }

    private void logError(String id, Failure<CristinProject> failure) {
        logger.error(String.format(ERROR_MESSAGE_FETCHING_CRISTIN_PROJECT_WITH_ID,
                                   id, failure.getException().getMessage()));
    }
```

axthosarouris: logError can also be used in enrichedProject
```
    private CristinProject enrichOneProject(String language, CristinProject project) {
        return attempt(() -> getProject(project.cristinProjectId, language))
            .toOptional(failure -> logError(project.cristinProjectId, failure))
            .orElse(project);
    }
```

axthosarouris (changes requested):
Unfinished review

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -66,12 +67,44 @@ public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetada
         projectsWrapper.setProcessingTime(calculateProcessingTime(startRequestTime, endRequestTime));
         // TODO: NP-2385: Use Link header / Pagination data from Cristin response in the next two values
         projectsWrapper.setFirstRecord(0);
-        projectsWrapper.setNextResults(null);
+        projectsWrapper.setNextResults("");
```
brinxmat, line 70: Is this really a case where we want an empty string? Here, I say we don't.

src/main/java/no/unit/nva/cristin/projects/RequestUtils.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+import java.util.Optional;
+import nva.commons.apigateway.RequestInfo;
+
+public class RequestUtils {
+
+    protected static final String LANGUAGE_QUERY_PARAMETER = "language";
+    protected static final String LANGUAGE_INVALID_ERROR_MESSAGE = "Parameter 'language' has invalid value";
+    protected static final String DEFAULT_LANGUAGE_CODE = "nb";
+    private static final List<String> VALID_LANGUAGE_CODES = Arrays.asList("nb", "en");
```
brinxmat, line 14: VALID_LANGUAGE_CODES is a set, not a list. Use Set.of("en", "nb");

src/main/java/no/unit/nva/cristin/projects/RequestUtils.java
```diff
@@ -0,0 +1,30 @@
...
+
+    protected static String getValidLanguage(RequestInfo requestInfo) throws BadRequestException {
+        return Optional.of(getQueryParam(requestInfo, LANGUAGE_QUERY_PARAMETER)
+            .orElse(DEFAULT_LANGUAGE_CODE))
+            .filter(RequestUtils::isValidLanguage)
```
brinxmat, line 19: VALID_LANGUAGE_CODES::contains

src/main/java/no/unit/nva/cristin/projects/RequestUtils.java
```diff
@@ -0,0 +1,30 @@
+package no.unit.nva.cristin.projects;
+
+import static nva.commons.core.attempt.Try.attempt;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import nva.commons.apigateway.RequestInfo;
+
+public class RequestUtils {
```
brinxmat, line 9: Is this in reality part of a shared interface for Fetch(One)?CristinProject(s)?

src/main/java/no/unit/nva/cristin/projects/RequestUtils.java
brinxmat: In general, when order and presence of duplicates is unimportant use HashSet. It also has better performance for contains operations.

src/test/java/no/unit/nva/cristin/projects/CommonUtilsTest.java
```diff
@@ -0,0 +1,29 @@
+package no.unit.nva.cristin.projects;
+
+import static no.unit.nva.cristin.projects.CommonUtils.hasValidContent;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import java.util.HashMap;
+import no.unit.nva.cristin.projects.model.cristin.CristinProject;
+import org.junit.jupiter.api.Test;
+
+public class CommonUtilsTest {
```
brinxmat, line 10: I don't think there is any reason to have such tests; this is very close to the implementation and doesn't add to coverage unless the methods are untested from an API-unit-test. In the latter case, it would cover up potential issues in the code (the method was in fact untested from the API).

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -46,17 +47,18 @@
     private static final String LANGUAGE_KEY = "language";
     private static final String EMPTY_STRING = "";
     private static final String LANGUAGE_NB = "nb";
-    private static final String LANGUAGE_INVALID = "invalid";
+    private static final String INVALID_LANGUAGE_PARAM = "ru";
```
brinxmat, line 50: Amusing for personal reasons :D

src/test/resources/api_query_response_no_projects_found.json
```diff
@@ -0,0 +1,10 @@
+{
+  "@context": "https://example.org/search-api-context.json",
+  "id": "https://api.dev.nva.aws.unit.no/project/search?QUERY_PARAMS",
+  "size": 0,
+  "searchString": "title=reindeer",
+  "processingTime": 1000,
+  "firstRecord": 0,
+  "nextResults": "",
```
brinxmat, line 8: Next results is a URI, while URI creation from an empty string is valid, it creates a null object, which I think would be misleading from an interface perspective, delegating discovery of null objects to the consumer.

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -46,17 +47,18 @@
...
 
     private static final String ALLOW_ALL_ORIGIN = "*";
     private static final String API_RESPONSE_NON_ENRICHED_PROJECTS_JSON = "api_response_non_enriched_projects.json";
     private static final String API_RESPONSE_ONE_CRISTIN_PROJECT_TO_NVA_PROJECT_JSON =
         "api_response_one_cristin_project_to_nva_project.json";
     private static final String CRISTIN_GET_PROJECT_RESPONSE = "cristinGetProjectResponse.json";
-    private static final ObjectMapper OBJECT_MAPPER = JsonUtils.objectMapper;
+    private static final String API_QUERY_RESPONSE_NO_PROJECTS_FOUND_JSON = "api_query_response_no_projects_found.json";
     private CristinApiClient cristinApiClientStub;
     private Environment environment = new Environment();
```
brinxmat, line 63: Make environment final

brinxmat (changes requested):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -153,4 +187,12 @@ private CristinProject enrichOneProject(String language, CristinProject project)
                 return project;
             });
     }
-}
\ No newline at end of file
+
+    @JsonInclude(NON_NULL)
+    private static class EmptyNvaProject extends NvaProject {
```
axthosarouris, line 192: put it on separate class.

src/main/java/no/unit/nva/cristin/projects/FetchOneCristinProject.java
```diff
@@ -0,0 +1,63 @@
...
+    public FetchOneCristinProject(CristinApiClient cristinApiClient, Environment environment) {
+        super(Void.class, environment);
+        this.cristinApiClient = cristinApiClient;
+    }
+
+    @Override
+    protected NvaProject processInput(Void input, RequestInfo requestInfo, Context context)
+        throws ApiGatewayException {
+
+        String language = getValidLanguage(requestInfo);
```
axthosarouris, line 42: Rename RequestUtils to SomeAbstractHandler that extends ApiGatewayhandler and add the  requestUtils methods there.

src/main/java/no/unit/nva/cristin/projects/RequestUtils.java
```diff
@@ -0,0 +1,30 @@
...
+
+    protected static String getValidLanguage(RequestInfo requestInfo) throws BadRequestException {
+        return Optional.of(getQueryParam(requestInfo, LANGUAGE_QUERY_PARAMETER)
+            .orElse(DEFAULT_LANGUAGE_CODE))
+            .filter(RequestUtils::isValidLanguage)
+            .orElseThrow(() -> new BadRequestException(LANGUAGE_INVALID_ERROR_MESSAGE));
+    }
+
+    protected static Optional<String> getQueryParam(RequestInfo requestInfo, String queryParameter) {
+        return attempt(() -> requestInfo.getQueryParameter(queryParameter)).toOptional();
```
axthosarouris, line 24: maybe we should make a case for adding this functionality to commons

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Test
+    void handlerReturnsNvaProjectFromTransformedCristinProjectWhenIdIsFound() throws Exception {
+        GatewayResponse<NvaProject> response = sendQueryWithId(DEFAULT_ID);
+        var expected = getReader(API_RESPONSE_ONE_PROJECT_JSON);
+        assertEquals(OBJECT_MAPPER.readTree(expected), OBJECT_MAPPER.readTree(response.getBody()));
+    }
+
+    @Test
+    void handlerReturnsEmptyJsonWhenFetchFromBackendFails() throws Exception {
```
axthosarouris, line 85: @brinxmat  Should we maybe return BadGateway here?

template.yaml
```diff
@@ -76,6 +76,24 @@ Resources:
               - method.request.querystring.language
               - method.request.querystring.title
 
+  NvaCristinOneProjectFunction:
```
axthosarouris, line 79: Curious to see what kind of OpenApi documentation will come up from this setting. We should test it and see if we need to change it in a subsequent PR

axthosarouris (changes requested):

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
brinxmat: If it fails, it is an error

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: It will come a value here in a later issue, when we have implemented logic for reading headers from backend call. Do you want me to use null for now until implementation?

brinxmat: Fair enough, I won't be pedantic :D

src/main/java/no/unit/nva/cristin/projects/model/cristin/CristinProject.java
```diff
@@ -1,27 +1,143 @@
...
+
+    public void setLastModified(Map<String, String> lastModified) {
+        this.lastModified = lastModified;
+    }
+
+    public CristinOrganization getCoordinatingInstitution() {
+        return coordinatingInstitution;
+    }
+
+    public void setCoordinatingInstitution(
```
brinxmat, line 115: Fix weird newlines in a couple of method defs.

src/main/java/no/unit/nva/cristin/projects/model/nva/EmptyNvaProject.java
```diff
@@ -0,0 +1,12 @@
+package no.unit.nva.cristin.projects.model.nva;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(NON_NULL)
+public class EmptyNvaProject extends NvaProject {
```
brinxmat, line 7: Not sure about this…should probably have a discussion (I probably don't understand)

brinxmat (changes requested):

axthosarouris (approved):

brinxmat (approved):

PR #31, opened 2021-04-12
Np 2424 map to empty instead of null

torarnet:


src/main/java/no/unit/nva/cristin/projects/NvaProjectBuilder.java
```diff
@@ -101,12 +103,12 @@ private String extractMainTitle() {
             .filter(titles -> titles.keySet().remove(cristinProject.getMainLanguage()))
             .filter(remainingTitles -> !remainingTitles.isEmpty())
             .map(Collections::singletonList)
-            .orElse(null);
+            .orElse(new ArrayList<>());
```
axthosarouris, line 106: `Collections.emptyList()`

axthosarouris (dismissed):

src/main/java/no/unit/nva/cristin/projects/model/nva/NvaProject.java
```diff
@@ -128,6 +130,14 @@ public void setEndDate(Instant endDate) {
         this.endDate = endDate;
     }
 
+    public List<Object> getGrant() {
```
brinxmat, line 133: Is the type of grant really Object?

src/test/resources/api_response_get_project_with_missing_fields.json
```diff
@@ -10,6 +10,9 @@
   ],
   "title": "Example Title",
   "language": "https://lexvo.org/id/iso639-3/nno",
+  "alternativeTitles": [],
   "startDate": "2018-11-01T00:00:00Z",
-  "endDate": "2021-05-01T00:00:00Z"
+  "endDate": "2021-05-01T00:00:00Z",
+  "grant": [],
```
brinxmat, line 16: Did you update the @context to ensure that properties that are collections are annotated @container: @set?

brinxmat (changes requested):

src/main/java/no/unit/nva/cristin/projects/model/nva/NvaProject.java
torarnet: It is just a placeholder until we actually implement this field in later issue

src/test/resources/api_response_get_project_with_missing_fields.json
torarnet: I did not know that was part of the specification. We need to discuss that

torarnet: After discussion: This will be solved in a later issue. 

Another conclusion: We will make some changes by using plural names for lists and send values with no content as null instead of not sending at all. This makes it better for clients

axthosarouris (approved):

brinxmat (approved):

PR #33, opened 2021-04-27
Np 2439 replace url openstream

torarnet:


src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -37,11 +38,51 @@
...
+        "Your request cannot be processed at this time because of an internal server error";
+    private static final HttpClient client = HttpClient.newHttpClient();
+    private static final int STATUS_CODE_NOT_FOUND = 404;
+    private static final int STATUS_CODE_START_OF_ERROR_CODES_RANGE = 299;
+    private static final String CRISTIN_PROJECT_MATCHING_ID_IS_NOT_VALID =
+        "Project matching id %s does not have valid data";
 
     protected static <T> T fromJson(InputStreamReader reader, Class<T> classOfT) throws IOException {
         return OBJECT_MAPPER.readValue(reader, classOfT);
     }
```
axthosarouris, line 53: Do we need both these methods?

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -37,11 +38,51 @@
...
+     * @throws BadGatewayException          when there is a problem with fetch from backend
+     * @throws InternalServerErrorException when there is an unforeseen error
+     * @throws NotFoundException            when upstream returns 404
+     */
+    public NvaProject queryOneCristinProjectUsingIdIntoNvaProject(String id, String language)
+        throws BadGatewayException, InternalServerErrorException, NotFoundException {
+
+        CristinProject cristinProject = attemptToGetCristinProject(id, language);
+
+        if (cristinProject == null || !cristinProject.hasValidContent()) {
```
axthosarouris, line 75: the team style is to use the isNull/nonNull methods

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -37,11 +38,51 @@
...
+            logger.warn(String.format(CRISTIN_PROJECT_MATCHING_ID_IS_NOT_VALID, id));
+            throw new BadGatewayException(String.format(CRISTIN_PROJECT_MATCHING_ID_IS_NOT_VALID, id));
+        }
+
+        NvaProject nvaProject = new NvaProjectBuilder(cristinProject).build();
+        nvaProject.setContext(Constants.PROJECT_LOOKUP_CONTEXT_URL);
```
axthosarouris, line 81: Since we are using a builder, add the context in the builder as well and then you can rewrite the whole thing as follows:

```
 public NvaProject queryOneCristinProjectUsingIdIntoNvaProject(String id, String language)
        throws BadGatewayException, InternalServerErrorException, NotFoundException {

      return Optional.ofNullable(attemptToGetCristinProject(id, language))
            .filter(CristinProject::hasValidContent)
            .map(NvaProjectBuilder::new)
            .map(builder->builder.withContext(Constants.PROJECT_LOOKUP_CONTEXT_URL))
            .map(NvaProjectBuilder::build)
            .orElseThrow(()->projectHasNotValidContent(id));
      
    }

    private BadGatewayException projectHasNotValidContent(String id) {
        logger.warn(String.format(CRISTIN_PROJECT_MATCHING_ID_IS_NOT_VALID, id));
        return new BadGatewayException(String.format(CRISTIN_PROJECT_MATCHING_ID_IS_NOT_VALID, id));
    }
```

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -137,43 +165,53 @@ private CristinProject attemptToGetCristinProject(String id, String language) th
...
+        HttpRequest httpRequest = HttpRequest.newBuilder(uri).build();
+
+        return attempt(() -> client.send(httpRequest, HttpResponse.BodyHandlers.ofInputStream())).orElseThrow();
     }
 
-    @JacocoGenerated
-    protected InputStreamReader fetchGetResult(URL url) throws IOException {
-        return new InputStreamReader(url.openStream());
+    protected URI generateGetProjectUri(String id, String language) throws URISyntaxException {
+        return new URIBuilder() // TODO: Replace URIBuilder() with only URI and put logic in UriUtils?
```
axthosarouris, line 183: Suggestion:

```
EMPTY_FRAGMENT = null;
 protected URI generateQueryProjectsUrl(Map<String, String> parameters) throws URISyntaxException {

        String query = queryParameters(parameters);
        return new URI(HTTPS, CRISTIN_API_HOST, CRISTIN_API_PROJECTS_PATH, query, EMPTY_FRAGMENT);
    }

UriUtis:
public static String queryParameters(Map<String, String> queryParameters) {
        return Optional.ofNullable(queryParameters)
                   .map(UriUtils::formatQueryParameters)
                   .orElse(null);
    }

    private static String formatQueryParameters(Map<String, String> queryParams) {
        return queryParams.entrySet().stream()
                   .map(entry -> String.format("%s=%s", entry.getKey(), entry.getValue()))
                   .collect(Collectors.joining("&"));
    }
```

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -137,43 +165,53 @@ private CristinProject attemptToGetCristinProject(String id, String language) th
...
+    protected URI generateQueryProjectsUrl(Map<String, String> parameters) throws URISyntaxException {
         URIBuilder uri = new URIBuilder()
             .setScheme(HTTPS)
             .setHost(CRISTIN_API_HOST)
             .setPath(CRISTIN_API_PROJECTS_PATH);
         if (parameters != null) {
             parameters.keySet().forEach(s -> uri.addParameter(s, parameters.get(s)));
         }
-        return uri.build().toURL();
+        return uri.build();
```
axthosarouris, line 199: ditto

src/main/java/no/unit/nva/cristin/projects/InternalServerErrorException.java
```diff
@@ -0,0 +1,18 @@
+package no.unit.nva.cristin.projects;
+
+import java.net.HttpURLConnection;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+import nva.commons.core.JacocoGenerated;
+
+@JacocoGenerated
+public class InternalServerErrorException extends ApiGatewayException {
```
axthosarouris, line 8: This should be in commons.

src/main/java/no/unit/nva/cristin/projects/NotFoundException.java
```diff
@@ -0,0 +1,18 @@
+package no.unit.nva.cristin.projects;
+
+import java.net.HttpURLConnection;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+import nva.commons.core.JacocoGenerated;
+
+@JacocoGenerated
+public class NotFoundException extends ApiGatewayException {
```
axthosarouris, line 8: This is already in commons

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
```diff
@@ -93,22 +106,28 @@ void handlerReturnsBadGatewayExceptionWhenFetchFromBackendFails() throws Excepti
...
-        assertEquals(APPLICATION_PROBLEM_JSON, response.getHeaders().get(HttpHeaders.CONTENT_TYPE));
+        assertEquals(HttpURLConnection.HTTP_INTERNAL_ERROR, nextResponse.getStatusCode());
+        assertEquals(APPLICATION_PROBLEM_JSON, nextResponse.getHeaders().get(HttpHeaders.CONTENT_TYPE));
     }
 
     @Test
     void handlerReturnsNvaProjectWithoutParticipantsAndCoordinatingInstitutionIfTheyAreMissingFromBackend()
         throws Exception {
 
         cristinApiClientStub = spy(cristinApiClientStub);
```
axthosarouris, line 127: Although it is not part of this PR,  I think what we should be mocking is the httpclient. But we should create a task for this for doing it later

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
brinxmat: The check-null-check-value pattern for me typically insinuates that the preceding call should return optional and the body of the if is an Optional::orElseThrow.

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -37,9 +40,45 @@
...
+     * @param language Language used for some properties in Cristin API response
+     * @return a NvaProject filled with one transformed Cristin Project
+     * @throws BadGatewayException          when there is a problem with fetch from backend
+     * @throws InternalServerErrorException when there is an unforeseen error
+     * @throws NotFoundException            when upstream returns 404
+     */
+    public NvaProject queryOneCristinProjectUsingIdIntoNvaProject(String id, String language)
+        throws BadGatewayException, InternalServerErrorException, NotFoundException {
+
+        return Optional.ofNullable(attemptToGetCristinProject(id, language))
```
brinxmat, line 67: I wonder if it isn't the case that attemptToGetCristinProject should return the optional since it makes the contract explicit? It *may* be null.

src/main/java/no/unit/nva/cristin/projects/UriUtils.java
```diff
@@ -2,10 +2,31 @@
...
+    /**
+     * Creates as string from a map containing a formatted query parameters string to be put on end of an url.
+     *
+     * @param queryParameters The query params
+     * @return a String with query parameters formatted to be placed at end of url
+     */
+    public static String queryParameters(Map<String, String> queryParameters) {
+        return Optional.ofNullable(queryParameters)
+            .map(UriUtils::formatQueryParameters)
+            .orElse(null);
```
brinxmat, line 24: Call this EMPTY_QUERY_PARAMETERS_FOR_URI_CONSTRUCTOR as it isn't clear why we would return null here rather than say Optional<String>.

src/main/java/no/unit/nva/cristin/projects/UriUtils.java
```diff
@@ -2,10 +2,31 @@
...
+    }
+
+    private static String formatQueryParameters(Map<String, String> queryParams) {
+        return queryParams.entrySet().stream()
+            .map(entry -> String.format("%s=%s", entry.getKey(), entry.getValue()))
```
brinxmat, line 29: Constant from "%s=%s" PARAMERTER_KEY_VALUE_PAIR_TEMPLATE

src/main/java/no/unit/nva/cristin/projects/UriUtils.java
```diff
@@ -2,10 +2,31 @@
...
+            .collect(Collectors.joining("&"));
```
brinxmat, line 30: & = PARAMETER_DELIMITER

brinxmat (changes requested):

build.gradle
```diff
@@ -13,12 +13,11 @@ targetCompatibility = JavaVersion.VERSION_11 // bytecode target version
```
axthosarouris, line 13: implementation 'org.apache.httpcomponents:httpclient:4.5.13' should be from the  dependencies.

axthosarouris (dismissed):

brinxmat (approved):

axthosarouris (approved):

PR #36, opened 2021-04-29
Np 2538 move transformation logic

torarnet:
A small PR but there were no other modeling logic in CristinApiClient

### DIFF
--- a/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
+++ b/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
@@ -7,7 +7,7 @@ import static no.unit.nva.cristin.projects.Constants.CRISTIN_LANGUAGE_PARAM;
 import static no.unit.nva.cristin.projects.Constants.LANGUAGE;
 import static no.unit.nva.cristin.projects.Constants.OBJECT_MAPPER;
 import static no.unit.nva.cristin.projects.Constants.TITLE;
-import static no.unit.nva.cristin.projects.NvaProjectBuilder.nvaProjectsFromCristinProjects;
+import static no.unit.nva.cristin.projects.NvaProjectBuilder.transformToNvaProject;
 import static no.unit.nva.cristin.projects.UriUtils.buildUri;
 import static no.unit.nva.cristin.projects.UriUtils.queryParameters;
 import static nva.commons.core.attempt.Try.attempt;
@@ -122,7 +122,7 @@ public class CristinApiClient {
         // TODO: NP-2385: Use Link header / Pagination data from Cristin response in the next two values
         projectsWrapper.setFirstRecord(0);
         projectsWrapper.setNextResults(null); // TODO: Change to URI
-        projectsWrapper.setHits(nvaProjectsFromCristinProjects(enrichedProjects));
+        projectsWrapper.setHits(transformToNvaProject(enrichedProjects));
 
         return projectsWrapper;
     }
--- a/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
+++ b/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
@@ -7,6 +7,7 @@ import static no.unit.nva.cristin.projects.Constants.CRISTIN_LANGUAGE_PARAM;
 import static no.unit.nva.cristin.projects.Constants.LANGUAGE;
 import static no.unit.nva.cristin.projects.Constants.OBJECT_MAPPER;
 import static no.unit.nva.cristin.projects.Constants.TITLE;
+import static no.unit.nva.cristin.projects.NvaProjectBuilder.nvaProjectsFromCristinProjects;
 import static no.unit.nva.cristin.projects.UriUtils.buildUri;
 import static no.unit.nva.cristin.projects.UriUtils.queryParameters;
 import static nva.commons.core.attempt.Try.attempt;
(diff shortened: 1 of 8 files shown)


src/main/java/no/unit/nva/cristin/projects/NvaProjectBuilder.java
```diff
@@ -115,4 +115,17 @@ public NvaProjectBuilder withContext(String context) {
...
         return this;
     }
+
+    /**
+     * Transforms a list of Cristin projects to a list of Nva projects.
+     *
+     * @param cristinProjects The list of Cristin projects to be transformed
+     * @return a List filled with Nva projects transformed from Cristin projects
+     */
+    public static List<NvaProject> nvaProjectsFromCristinProjects(List<CristinProject> cristinProjects) {
```
brinxmat, line 125: transformToNvaProject

brinxmat (changes requested):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -120,7 +122,7 @@ public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetada
         // TODO: NP-2385: Use Link header / Pagination data from Cristin response in the next two values
         projectsWrapper.setFirstRecord(0);
         projectsWrapper.setNextResults(null); // TODO: Change to URI
-        projectsWrapper.setHits(transformCristinProjectsToNvaProjects(enrichedProjects));
+        projectsWrapper.setHits(transformToNvaProject(enrichedProjects));
```
axthosarouris, line 125: Suggestion:
In  CrstinApiClient:
```
  public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetadata(
        Map<String, String> requestQueryParams)
        throws IOException, URISyntaxException {

        long startRequestTime = System.currentTimeMillis();

        List<NvaProject> nvaProjects = importProjectsFromCristin(requestQueryParams);
        ProjectsWrapper projectsWrapper = new ProjectsWrapper();

        String requestQueryParamsAsString = String.format(TITLE_AND_LANGUAGE_QUERY_PARAM_PLACEHOLDER,
            requestQueryParams.get(TITLE), requestQueryParams.get(LANGUAGE));

        projectsWrapper.setId(buildUri(BASE_URL, QUESTION_MARK + requestQueryParamsAsString));
        projectsWrapper.setSize(0); // TODO: NP-2385: X-Total-Count header from Cristin response
        projectsWrapper.setSearchString(requestQueryParamsAsString);
        projectsWrapper.setProcessingTime(calculateProcessingTime(startRequestTime, endRequestTime));
        // TODO: NP-2385: Use Link header / Pagination data from Cristin response in the next two values
        projectsWrapper.setFirstRecord(0);
        projectsWrapper.setNextResults(null); // TODO: Change to URI
        projectsWrapper.setHits(nvaProjects);

        return projectsWrapper;
    }

    private List<NvaProject> importProjectsFromCristin(Map<String, String> requestQueryParams) throws IOException, URISyntaxException {
        return queryAndEnrichProjects(requestQueryParams)
            .filter(CristinProject::hasValidContent)
                  .map(CristinProject::toNvaProject)
                  .collect(Collectors.toList());
    }
```
In CristinProject:

```
  public NvaProject toNvaProject(){
        return new NvaProjectBuilder(this).build();
    }
```

axthosarouris: I notice that we collect a stream and we stream it again so basically we do this:
```
someList.stream().map(x->toY(x)).collect().stream().map(y->toZ(z)).collect().stream()...
```
If a stream is to be processed once, it is better to collect only at the end I think.  I find no harm to break the processing of stream in several parts but we do not need to collect and stream again if there is no reason for that.
In my above suggestion `queryAndEnrichProjects` returns a Stream.

brinxmat (dismissed):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -105,26 +106,37 @@ public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetada
...
         // TODO: NP-2385: Use Link header / Pagination data from Cristin response in the next two values
         projectsWrapper.setFirstRecord(0);
         projectsWrapper.setNextResults(null); // TODO: Change to URI
-        projectsWrapper.setHits(transformCristinProjectsToNvaProjects(enrichedProjects));
+        projectsWrapper.setHits(nvaProjects);
 
         return projectsWrapper;
     }
 
+    private String titleAndLanguageQueryString(Map<String, String> requestQueryParams) {
```
torarnet, line 126: ould have reused UriUtils.queryParameters but i needed to maintain order of entries for unit testing

axthosarouris: I find this situation quite fragile. Is there a reason why the query parameters should be in a specific order? I think the solution to the problem is to normalize the URI by ordering the query parameters alphabetically. 

The method UriUtils.formatQueryParameters should then change to 
```
  private static String formatQueryParameters(Map<String, String> queryParams) {
        return queryParams.entrySet().stream()
                   .sorted(Entry.comparingByKey())
                   .map(entry -> String.format(PARAMETER_KEY_VALUE_PAIR_TEMPLATE, entry.getKey(), entry.getValue()))
                   .collect(Collectors.joining(PARAMETER_DELIMITER));
    }
```
The resource files should be corrected as well.

axthosarouris: About URI normalization: https://en.wikipedia.org/wiki/URI_normalization

axthosarouris (changes requested):

axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/UriUtils.java
```diff
@@ -31,6 +32,7 @@ public static String queryParameters(Map<String, String> queryParameters) {
 
     private static String formatQueryParameters(Map<String, String> queryParams) {
         return queryParams.entrySet().stream()
+            .sorted(Entry.comparingByKey())
```
brinxmat, line 35: Since we're sorting, is it better to change the argument of the method to a SortedMap and pass a TreeMap that orders the keys by natural key order?

brinxmat (changes requested):

PR #39, opened 2021-05-03
Np 2437 handle exceptions differently

torarnet:


src/main/java/no/unit/nva/cristin/projects/Constants.java
```diff
@@ -22,4 +22,5 @@
     public static final String CRISTIN_LANGUAGE_PARAM = "lang";
     public static final String TITLE = "title";
     public static final String LANGUAGE = "language";
+    public static final String QUESTION_MARK = "?";
```
axthosarouris, line 25: PARAMETER_ARRAY_PREFIX. If it is to name it question mark, then better to use the actual question mark.

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
```diff
@@ -109,15 +110,15 @@ void handlerReturnsBadGatewayWhenBackendThrowsIoException() throws Exception {
...
     @Test
-    void handlerReturnsServerErrorExceptionWhenBackendThrowsUriSyntaxException() throws Exception {
+    void handlerReturnsServerErrorExceptionWhenBackendThrowsGenericException() throws Exception {
         cristinApiClientStub = spy(cristinApiClientStub);
 
-        doThrow(URISyntaxException.class).when(cristinApiClientStub).getProject(any(), any());
+        doThrow(RuntimeException.class).when(cristinApiClientStub).getProject(any(), any());
         handler = new FetchOneCristinProject(cristinApiClientStub, environment);
-        GatewayResponse<NvaProject> nextResponse = sendQueryWithId(DEFAULT_ID);
+        GatewayResponse<NvaProject> response = sendQueryWithId(DEFAULT_ID);
```
axthosarouris, line 118: This is a GatewayResponse<Problem>  I think

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
```diff
@@ -154,6 +155,45 @@ void getsCorrectUriWhenCallingGetProjectUriBuilder() throws Exception {
...
     }
 
+    @Test
+    void handlerThrowsBadGatewayExceptionWhenThereIsThrownExceptionWhenReadingFromJson() throws Exception {
+        cristinApiClientStub = spy(cristinApiClientStub);
+
+        doReturn(new HttpResponseStub(IoUtils.stringToStream("")))
+            .when(cristinApiClientStub).fetchGetResult(any(URI.class));
+        handler = new FetchOneCristinProject(cristinApiClientStub, environment);
+        GatewayResponse<NvaProject> response = sendQueryWithId(DEFAULT_ID);
```
axthosarouris, line 165: same here.

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
```diff
@@ -109,15 +110,15 @@ void handlerReturnsBadGatewayWhenBackendThrowsIoException() throws Exception {
...
 
-        doThrow(URISyntaxException.class).when(cristinApiClientStub).getProject(any(), any());
+        doThrow(RuntimeException.class).when(cristinApiClientStub).getProject(any(), any());
         handler = new FetchOneCristinProject(cristinApiClientStub, environment);
-        GatewayResponse<NvaProject> nextResponse = sendQueryWithId(DEFAULT_ID);
+        GatewayResponse<NvaProject> response = sendQueryWithId(DEFAULT_ID);
 
-        assertEquals(HttpURLConnection.HTTP_INTERNAL_ERROR, nextResponse.getStatusCode());
-        assertEquals(APPLICATION_PROBLEM_JSON, nextResponse.getHeaders().get(HttpHeaders.CONTENT_TYPE));
+        assertEquals(HttpURLConnection.HTTP_INTERNAL_ERROR, response.getStatusCode());
```
axthosarouris, line 120: What is the expected message to the api-client?

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
axthosarouris: what is the expected message to the api-client?

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
```diff
@@ -154,6 +155,45 @@ void getsCorrectUriWhenCallingGetProjectUriBuilder() throws Exception {
...
+    void handlerReturnsInternalErrorWhenUriCreationFails() throws Exception {
+        cristinApiClientStub = spy(cristinApiClientStub);
+
+        doThrow(URISyntaxException.class).when(cristinApiClientStub).generateGetProjectUri(any(), any());
+
+        handler = new FetchOneCristinProject(cristinApiClientStub, environment);
+        GatewayResponse<NvaProject> response = sendQueryWithId(DEFAULT_ID);
+
+        assertEquals(HttpURLConnection.HTTP_INTERNAL_ERROR, response.getStatusCode());
+        assertEquals(APPLICATION_PROBLEM_JSON, response.getHeaders().get(HttpHeaders.CONTENT_TYPE));
```
axthosarouris, line 181: ditto

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
```diff
@@ -154,6 +155,45 @@ void getsCorrectUriWhenCallingGetProjectUriBuilder() throws Exception {
...
+    void handlerThrowsInternalErrorWhenHttpStatusCodeIsSomeUnexpectedValue() throws Exception {
+        cristinApiClientStub = spy(cristinApiClientStub);
+
+        doReturn(new HttpResponseStub(IoUtils.stringToStream(""), 418))
+            .when(cristinApiClientStub).fetchGetResult(any(URI.class));
+        handler = new FetchOneCristinProject(cristinApiClientStub, environment);
+        GatewayResponse<NvaProject> response = sendQueryWithId(DEFAULT_ID);
+
+        assertEquals(HttpURLConnection.HTTP_INTERNAL_ERROR, response.getStatusCode());
+        assertEquals(APPLICATION_PROBLEM_JSON, response.getHeaders().get(HttpHeaders.CONTENT_TYPE));
```
axthosarouris, line 194: ditto

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -232,8 +226,54 @@ void handlerReturnsProjectsWrapperWithAllMetadataButEmptyHitsArrayWhenNoMatchesA
...
+
+        handler = new FetchCristinProjects(cristinApiClientStub, environment);
+        GatewayResponse<ProjectsWrapper> response = sendDefaultQuery();
+
+        assertEquals(HttpURLConnection.HTTP_INTERNAL_ERROR, response.getStatusCode());
+        assertEquals(APPLICATION_PROBLEM_JSON, response.getHeaders().get(HttpHeaders.CONTENT_TYPE));
+    }
+
+    @Test
+    void cristinQueryClassBuildsCorrectUriWhenSuppliedWithParameters() throws Exception {
```
axthosarouris, line 272: This belongs to the CristinQueryTest class, and the test name should follow the pattern.

src/main/java/no/unit/nva/cristin/projects/CristinQuery.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final String HTTPS = "https";
+    private static final String CRISTIN_API_PROJECTS_PATH = "/v2/projects/";
+    private static final String EMPTY_FRAGMENT = null;
+
+    private final transient Map<String, String> cristinQueryParameters;
+
+    /**
+     * Creates a object used to generate URI to connect to Cristin Projects.
+     */
+    public CristinQuery() {
```
axthosarouris, line 27: The constructor should be private since you only use the public static method. It is not a general rule but I do not see why the class needs to have two entry points. But this is a minor comment

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -84,47 +84,47 @@ void handlerReturnsExpectedBodyWhenRequestInputIsValid(String expected) throws I
...
 
     @Test
-    public void handlerIgnoresErrorsWhenTryingToEnrichProjectInformation() throws Exception {
+    void handlerThrowsInternalErrorWhenQueryingProjectsFails() throws Exception {
         cristinApiClientStub = spy(cristinApiClientStub);
-        doThrow(new IOException()).when(cristinApiClientStub).getProject(any(), any());
+        doThrow(new RuntimeException()).when(cristinApiClientStub).queryAndEnrichProjects(any());
         handler = new FetchCristinProjects(cristinApiClientStub, environment);
 
         GatewayResponse<ProjectsWrapper> response = sendDefaultQuery();
```
axthosarouris, line 110: when throwing exceptions we want to be sure that we return Problems. and we should also check the message to the API-client, to be sure that we are not exposing too much information.

axthosarouris (changes requested):

src/main/java/no/unit/nva/cristin/projects/Constants.java
brinxmat: I looked in the URI spec, it is actually called "question mark"

src/main/java/no/unit/nva/cristin/projects/CristinQuery.java
```diff
@@ -0,0 +1,88 @@
+package no.unit.nva.cristin.projects;
+
+import static no.unit.nva.cristin.projects.Constants.CRISTIN_API_HOST;
+import static no.unit.nva.cristin.projects.UriUtils.queryParameters;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class CristinQuery {
```
brinxmat, line 10: I think this class is actually CristinQueryUriBuilder(String id).

brinxmat: Address this in a new PR

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -84,47 +84,47 @@ void handlerReturnsExpectedBodyWhenRequestInputIsValid(String expected) throws I
...
         cristinApiClientStub = spy(cristinApiClientStub);
-        doThrow(new IOException()).when(cristinApiClientStub).queryAndEnrichProjects(any());
+        doThrow(BadGatewayException.class).when(cristinApiClientStub).getProject(any(), any());
         handler = new FetchCristinProjects(cristinApiClientStub, environment);
-
         GatewayResponse<ProjectsWrapper> response = sendDefaultQuery();
-
-        assertEquals(HttpURLConnection.HTTP_INTERNAL_ERROR, response.getStatusCode());
-        assertEquals(APPLICATION_PROBLEM_JSON, response.getHeaders().get(HttpHeaders.CONTENT_TYPE));
+        var expected = IoUtils.stringFromResources(Path.of(API_RESPONSE_NON_ENRICHED_PROJECTS_JSON));
```
brinxmat, line 122: I don't disagree with the use of var, but I guess we should be consistent across the project.

brinxmat (changes requested):

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
torarnet: you mean i should call response for gatewayResponse?

src/main/java/no/unit/nva/cristin/projects/CristinQuery.java
torarnet: I use constructor for search and static method for lookup. Should i change that?

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
torarnet: Should i add an expected message to HTTP_INTERNAL_ERROR exceptions only or add them to all relevant tests?

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
torarnet: Can you elaborate on test name? Something like `buildsCorrectQueryUriWithParametersWhenUsingCristinQueryClass`?

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
torarnet: Should i do something like this: `assertTrue(response.getBody().contains(ERROR_MESSAGE_BACKEND_FETCH_FAILED));`
Or should i write a POJO and deserialize for then to read and compare the detail field? And should that be a separate PR?

src/main/java/no/unit/nva/cristin/projects/Constants.java
axthosarouris: My bad  then :)

src/main/java/no/unit/nva/cristin/projects/CristinQuery.java
axthosarouris: better create two static methods search and lookup?

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
axthosarouris: I thought we were already using zalando Problem. Just leave it as it is and it can be fixed in another PR

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
axthosarouris: buildRetursnUriWithCustomParameterValuesWhenCustomParameterValuesAreSupplied

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
axthosarouris: Sorry I mean Problem class, should be fixed in subsequent PR

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
axthosarouris: leave it to next PR

axthosarouris (commented):
About Zalando Problem: https://github.com/zalando/problem
Check its usage in publication-api or user-access-service or we can talk 

Requested Change: the test that should be in CristinQueryTest with updated name

axthosarouris (approved):

brinxmat (approved):

PR #41, opened 2021-05-06
NP-2536: Moved Http Response further up the method chain

torarnet:
Needs to be done to get access to the response headers for use in ProjectsWrapper

### DIFF
--- a/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
+++ b/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
@@ -63,25 +63,6 @@ public class CristinApiClient {
             .orElseThrow(() -> projectHasNotValidContent(id));
     }
 
-    protected List<CristinProject> queryProjects(Map<String, String> parameters) throws ApiGatewayException {
-        URI uri = attempt(() -> generateQueryProjectsUrl(parameters))
-            .toOptional(failure ->
-                logError(ERROR_MESSAGE_QUERY_WITH_PARAMS_FAILED, queryParameters(parameters), failure.getException()))
-            .orElseThrow();
-
-        HttpResponse<InputStream> response = fetchQueryResults(uri);
-
-        checkHttpStatusCode(
-            buildUri(BASE_URL, QUESTION_MARK + queryParameters(parameters)).toString(),
-            response.statusCode());
-
-        return asList(getDeserializedResponse(response, CristinProject[].class));
-    }
-
-    protected static <T> T fromJson(InputStream stream, Class<T> classOfT) throws IOException {
-        return OBJECT_MAPPER.readValue(stream, classOfT);
-    }
-
     /**
      * Creates a wrapper object containing Cristin Projects transformed to NvaProjects with additional metadata. Is used
      * for serialization to the client.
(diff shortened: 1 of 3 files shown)


src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -94,9 +75,15 @@ public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetada
         Map<String, String> requestQueryParams) throws ApiGatewayException {
 
         long startRequestTime = System.currentTimeMillis();
-        List<NvaProject> nvaProjects = importProjectsFromCristin(requestQueryParams);
+        HttpResponse<InputStream> response = queryProjects(requestQueryParams);
```
axthosarouris, line 78: Require explicitly that you have a UTF-8 encoded content by altering fetchGetResult to
```
   @JacocoGenerated
    protected HttpResponse<String> fetchGetResult(URI uri) {
        HttpRequest httpRequest = HttpRequest.newBuilder(uri).build();

        return attempt(() -> client.send(httpRequest, BodyHandlers.ofString(StandardCharsets.UTF_8))).orElseThrow();
    }
```
Either way or the other we require it implicitly when we use IoUtils.streamToString()

torarnet: You want me to change the response type body to String instead of InputStream? Do you want me to do it for both query/search and lookup? Should i do that in separate PR or in this one? Personally I think it should be its own PR with an optional issue

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -94,9 +75,15 @@ public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetada
...
+        List<CristinProject> enrichedProjectsFromResponse =
+            getEnrichedProjectsUsingQueryResponse(response, requestQueryParams.get(LANGUAGE));
+        List<NvaProject> nvaProjects = mapValidCristinProjectsToNvaProjects(enrichedProjectsFromResponse);
         long endRequestTime = System.currentTimeMillis();
 
+        // TODO: NP-2385
+        //HttpHeaders headers = response.headers();
+
         ProjectsWrapper projectsWrapper = new ProjectsWrapper();
```
janbj, line 87: This method is doing a lot, consider extracting the creation of the wrapper to a new method

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
axthosarouris: OK

axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -94,9 +75,15 @@ public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetada
         Map<String, String> requestQueryParams) throws ApiGatewayException {
 
         long startRequestTime = System.currentTimeMillis();
-        List<NvaProject> nvaProjects = importProjectsFromCristin(requestQueryParams);
+        HttpResponse<InputStream> response = queryProjects(requestQueryParams);
+        List<CristinProject> enrichedProjectsFromResponse =
```
janbj, line 79: What makes it an enriched project?

Maybe call this mapResponseToCristinProjects()

janbj (approved):

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: It is a separate issue to refactor building of a ProjectsWrapper object, so it will be addressed later

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: I have created issue for it

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
torarnet: Query of projects does not give us much data about each project, so we have to manually lookup each project from the query using id to get more fields

PR #43, opened 2021-05-07
NP-2539: ProjectsWrapper now sets fields internally from builder methods

torarnet:


src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
```diff
@@ -104,4 +110,42 @@ public void setNextResults(String nextResults) {
...
+    // TODO: NP-2385
+
+    /**
+     * Assigns values to some of the fields using supplied http headers.
+     *
+     * @param headers the headers
+     * @return ProjectsWrapper object with field values from http headers
+     */
+    public ProjectsWrapper usingHeaders(HttpHeaders headers) {
+        this.size = 0;
```
axthosarouris, line 136: General comment: I think this is ahead of time implementation. We are implementing a method that will be useful the feature and we will fill in the gaps, probably without testing the added functionality. If this method is not doing much, it should be removed and when the relevant feature is in progress then we should write the method.

brinxmat (approved):

brinxmat (approved):

PR #48, opened 2021-05-19
Np 2385 add cristin headers

torarnet:


torarnet:
This is the first PR for this issue. More will come later to create smaller PR instead of one large one

brinxmat (changes requested):

src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
```diff
@@ -112,33 +115,42 @@ public void setHits(List<NvaProject> hits) {
      * Assigns value to id using supplied query parameters.
      *
      * @param queryParams the query params
-     * @return ProjectsWrapper object with id value from query parameters
+     * @return ProjectsWrapper object with id and firstRecord values from query parameters
      */
     public ProjectsWrapper usingQueryParams(Map<String, String> queryParams) {
         this.id = idUriFromParams(queryParams);
```
axthosarouris, line 121: At line 121 we calculate the query URI again (the other method being `CristinApiClient::generateQueryProjectsUrl`).
Ideally `CristinApiClient::queryProjects` should return an object containing the response and the query URI so that we know which query produced the response instead of calculating again. As a workaround now we could re-use the same method (`CristinApiClient::generateQueryProjectsUrl`), so that we do not implement the same logic twice.

axthosarouris (changes requested):

src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
torarnet: line 121 is the Nva URI while CristinApiClient::generateQueryProjectsUrl is Cristin URI

src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
```diff
@@ -112,33 +115,42 @@ public void setHits(List<NvaProject> hits) {
...
+        this.firstRecord = firstRecordFromParams(queryParams);
```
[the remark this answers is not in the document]
torarnet: If you want to change the model I can update afterwards, then for noe just use suggestion of indexOfFirstEntryInPage

brinxmat (approved):

axthosarouris (approved):

PR #49, opened 2021-05-21
NP-2385: Refactor test code to be reused for testing link header later

torarnet:
Added previousResults field
Combined header and query params into single method because later logic for uri's need both
Changed type of previous and nextResults to URI
firstRecord will now not exceed the size of the result set
Useful if you ask for pagination which results in no hits

### DIFF
--- a/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
+++ b/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
@@ -87,8 +87,7 @@ public class CristinApiClient {
         long endRequestTime = System.currentTimeMillis();
 
         return new ProjectsWrapper()
-            .usingQueryParams(requestQueryParams)
-            .usingHeaders(response.headers())
+            .usingHeadersAndQueryParams(response.headers(), requestQueryParams)
             .withProcessingTime(calculateProcessingTime(startRequestTime, endRequestTime))
             .withHits(nvaProjects);
     }
--- a/src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
+++ b/src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
@@ -21,6 +21,7 @@ import static no.unit.nva.cristin.projects.JsonPropertyNames.SEARCH_STRING;
 import static no.unit.nva.cristin.projects.JsonPropertyNames.SIZE;
 import static no.unit.nva.cristin.projects.UriUtils.queryParameters;
 import static nva.commons.core.attempt.Try.attempt;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
@@ -38,6 +39,9 @@ import nva.commons.core.JacocoGenerated;
 @JsonPropertyOrder({CONTEXT, ID, SIZE, SEARCH_STRING, PROCESSING_TIME, FIRST_RECORD, NEXT_RESULTS, PREVIOUS_RESULTS,
     HITS})
 public class ProjectsWrapper {
+
+    @JsonIgnore
+    public static final int FIRST_RECORD_ZERO_WHEN_NO_HITS = 0;
 
     @JsonProperty("@context")
     private String context = PROJECT_SEARCH_CONTEXT_URL;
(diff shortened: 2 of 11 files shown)


src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
```diff
@@ -112,17 +124,27 @@ public void setHits(List<NvaProject> hits) {
...
+     * @param headers     the headers from response
+     * @param queryParams the query params from request
+     * @return ProjectsWrapper object with some of the field values set using the supplied parameters
      */
-    public ProjectsWrapper usingQueryParams(Map<String, String> queryParams) {
+    public ProjectsWrapper usingHeadersAndQueryParams(HttpHeaders headers, Map<String, String> queryParams) {
+        this.size = getSizeHeader(headers);
         this.id = idUriFromParams(queryParams);
         this.firstRecord = indexOfFirstEntryInPageCalculatedFromParams(queryParams);
+        setFirstRecordToZeroIfExceedsMaxSizeOfResultSet();
```
brinxmat, line 137: Probably simplify as:
```
        int firstRecord = indexOfFirstEntryInPageCalculatedFromParams(queryParams);
        this.firstRecord = (this.size > firstRecord) ? firstRecord : 0;
```

brinxmat (changes requested):

src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
torarnet: I made a method with a descriptive name to avoid commenting why we do it. Should i drop the method and just use the logic directly as you imply or keep the method and change the logic in that method?

brinxmat:
```
    public ProjectsWrapper usingHeadersAndQueryParams(HttpHeaders headers, Map<String, String> queryParams) {
        this.size = getSizeHeader(headers);
        this.id = idUriFromParams(queryParams);
        int firstRecord = indexOfFirstEntryInPageCalculatedFromParams(queryParams);
        this.firstRecord = pageIsEmpty(firstRecord) ? firstRecord : 0;

        return this;
    }

    private boolean pageIsEmpty(int firstRecord) {
        return this.size > firstRecord;
    }
    
```

src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -87,8 +87,7 @@ public ProjectsWrapper queryCristinProjectsIntoWrapperObjectWithAdditionalMetada
         long endRequestTime = System.currentTimeMillis();
 
         return new ProjectsWrapper()
-            .usingQueryParams(requestQueryParams)
-            .usingHeaders(response.headers())
+            .usingHeadersAndQueryParams(response.headers(), requestQueryParams)
```
axthosarouris, line 90: I think this is an indication of a problem in design. We are using a Builder like pattern, but we are mutating the object with every function. That is why we need to set the headers and query parameters at the same time.

src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
axthosarouris: I think this is a BadRequest. We set a parameter "size" to be less than the "firstRecord" and we get an OK and a possibly unexpected result.  If I say bring me the 12th item from a collection of 10 items the response should not be "here is the first one", but  "your request is invalid"

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -312,11 +303,16 @@ void handlerThrowsBadRequestWhenQueryHasInvalidPageParameter() throws Exception
     }
 
     @ParameterizedTest(name = "Handler returns firstRecord {0} when record has pagination {1} and is on page {2}")
-    @CsvSource({"1,200,1", "11,5,3", "226,25,10", "55,9,7"})
+    @CsvSource({"1,200,1", "11,5,3", "226,25,10", "55,9,7", "0,50,10"})
```
axthosarouris, line 306: I think the case (0,50,100) should be its own test

axthosarouris (changes requested):

src/main/java/no/unit/nva/cristin/projects/ProjectsWrapper.java
```diff
@@ -112,17 +126,30 @@ public void setHits(List<NvaProject> hits) {
...
+     * @throws BadRequestException if page requested is invalid
      */
-    public ProjectsWrapper usingQueryParams(Map<String, String> queryParams) {
+    public ProjectsWrapper usingHeadersAndQueryParams(HttpHeaders headers, Map<String, String> queryParams)
+        throws BadRequestException {
+
+        this.size = getSizeHeader(headers);
         this.id = idUriFromParams(queryParams);
-        this.firstRecord = indexOfFirstEntryInPageCalculatedFromParams(queryParams);
+        this.firstRecord = this.size > 0 ? indexOfFirstEntryInPageCalculatedFromParams(queryParams) : 0;
```
axthosarouris, line 141: can you create a constant for the second 0? because it in unclear if it is the first element of an array (sometimes the first element is the 0th element) of it indicates that the value is not set.

axthosarouris (approved):

brinxmat (approved):

PR #52, opened 2021-05-31
NP-2444: Add open api file

torarnet:


docs/projects-swagger.yaml
```diff
@@ -0,0 +1,416 @@
...
+      description: Returns a list of projects based on query parameters
+      operationId: ListProjects
+      parameters:
+        - name: title
+          in: query
+          description: Title search string
+          required: true
+          schema:
+            type: string
+            pattern: '\\w'
```
axthosarouris, line 44: No special characters in title? e.g. question marks?

docs/projects-swagger.yaml
```diff
@@ -0,0 +1,416 @@
...
+        previousResults:
+          type: string
+          pattern: 'https:\/\/.*$'
+          description: URI to the previous page in the query results
+        hits:
+          type: array
+          description: The list of projects returned by the query
+          items:
+            $ref: '#/components/schemas/OneProject'
+    OneProject:
```
axthosarouris, line 202: Maybe call it Project?

docs/projects-swagger.yaml
```diff
@@ -0,0 +1,416 @@
...
+                type: string
+                pattern: 'CristinIdentifier'
+                description: Name of identifier, always CristinIdentifier
+              value:
+                type: integer
+                format: int32
+                description: Cristin ID of project
+        title:
+          type: string
+          pattern: '\\w'
```
axthosarouris, line 229: Same comment as before for the special characters, Also doesn't `\w` mean one letter only? shouldn't it be '\w+` or `.+`?

docs/projects-swagger.yaml
```diff
@@ -0,0 +1,416 @@
...
+                properties:
+                  id:
+                    $ref: '#/components/schemas/Id'
+                  type:
+                    type: string
+                    pattern: 'Person'
+                    description: Type of identity. Always Person
+                  firstName:
+                    type: string
+                    pattern: '\\w'
```
axthosarouris, line 274: I think `\\w` means one character only

axthosarouris (changes requested):

axthosarouris (dismissed):

docs/projects-swagger.yaml
```diff
@@ -0,0 +1,416 @@
...
+        uri:
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NvaCristinOneProjectFunction.Arn}/invocations
+        httpMethod: POST
+        type: "AWS_PROXY"
+      tags:
+        - Project
+      summary: Find project by ID
+      description: Returns a single project matching id parameter
+      operationId: GetProjectById
+      parameters:
```
brinxmat, line 107: Does this have a query param for language? If yes, extract a Schema component for the language param and use that

brinxmat (changes requested):

docs/projects-swagger.yaml
```diff
@@ -0,0 +1,429 @@
...
+          description: URI to the previous page in the query results
+        hits:
+          type: array
+          description: The list of projects returned by the query
+          items:
+            $ref: '#/components/schemas/Project'
+    Project:
+      type: object
+      properties:
+        '@context':
```
torarnet, line 211: is there a way to remove context when adding this schema as part of List schema? We do not want to include context 2 times and i am trying to avoid duplicating all other fields of one project to match list of projects

template.yaml
```diff
@@ -45,6 +45,11 @@ Resources:
   NvaCristinProjectsFunctionApi:
     Type: AWS::Serverless::Api
     Properties:
+      DefinitionBody:
+        'Fn::Transform':
+          Name: AWS::Include
+          Parameters:
+            Location: ./docs/projects-swagger.yaml
```
janbj, line 52: Has this been tested? Do you need to include a S3 bucket location?

torarnet: As far as I know, I have no way to test a template. This code was an example given by @svenngjeto so I just have to assume it works. If you have some suggestions please let me know

brinxmat (approved):

template.yaml
janbj: Verified in CFN deploy.

PR #55, opened 2021-06-04
Np 2155 search grant

torarnet:
Client can now search with both grantid and title in `query` field. If searching with a numerical value it searches for grantid first and if no hits found, search for that same value in title. Title query is as before.

### DIFF
--- a/src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
+++ b/src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
@@ -26,6 +26,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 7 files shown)


src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -243,14 +251,27 @@ void getsCorrectUriWhenCallingQueryProjectsUriBuilder() throws Exception {
...
+    }
+
+    @Test
+    void getsCorrectUriWhenCallingQueryGrantIdUriBuilder() throws Exception {
+        Map<String, String> params = Map.of(
+            QUERY, GRANT_ID_EXAMPLE,
+            LANGUAGE, LANGUAGE_NB,
+            PAGE, FIRST_PAGE,
+            NUMBER_OF_RESULTS, DEFAULT_NUMBER_OF_RESULTS);
+        URI uri = new URI(CRISTIN_API_GRANT_ID_SEARCH_EXAMPLE_URI);
```
axthosarouris, line 264: expectedUri

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -243,14 +251,27 @@ void getsCorrectUriWhenCallingQueryProjectsUriBuilder() throws Exception {
             NUMBER_OF_RESULTS, DEFAULT_NUMBER_OF_RESULTS);
         URI uri = new URI(QUERY_CRISTIN_PROJECTS_EXAMPLE_URI);
 
-        assertEquals(uri, cristinApiClientStub.generateQueryProjectsUrl(params));
+        assertEquals(uri, cristinApiClientStub.generateQueryProjectsUrl(params, QUERY_USING_TITLE));
+    }
+
+    @Test
+    void getsCorrectUriWhenCallingQueryGrantIdUriBuilder() throws Exception {
```
axthosarouris, line 258: if this is a test for the stub it should be clarified in the test (i.e.. "stubGets..."). Maybe it would be better to be moved in own class

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -450,6 +472,62 @@ void handlerReturnsProjectsWrapperWithCorrectNextResultsAndPreviousResultsWhenLi
         assertEquals(expectedPrevious, actualPrevious);
     }
 
+    @Test
+    void handlerReturnsMatchingProjectsFromGrantIdSearchWhenSuppliedWithOnlyNumber() throws Exception {
```
axthosarouris, line 446: I am not sure what we are testing here. I changed the value "GRANT_ID_EXAMPLE" with a random string ("sdfsf") and the test passed

axthosarouris: I think the testing has become very complicated because we are testing both the client and the mocking server. 

I think the idea is that "if the query is a number search first for ids and if empty search for title", so basically we should set up the mock server to return empty in the first call and something in the second.

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -450,6 +472,62 @@ void handlerReturnsProjectsWrapperWithCorrectNextResultsAndPreviousResultsWhenLi
...
+    void handlerReturnsProjectsFromTitleSearchWhenSuppliedWithQueryStringIncludingNumber() throws Exception {
+        cristinApiClientStub = spy(cristinApiClientStub);
+
+        doThrow(RuntimeException.class)
+            .when(cristinApiClientStub).queryProjects(any(), eq(QUERY_USING_GRANT_ID));
+
+        doReturn(new HttpResponseStub(CRISTIN_QUERY_PROJECTS_RESPONSE_JSON_FILE))
+            .when(cristinApiClientStub).queryProjects(any(), eq(QUERY_USING_TITLE));
+
+        InputStream input = requestWithQueryParameters(Map.of(QUERY, GRANT_ID_EXAMPLE + RANDOM_TITLE));
```
axthosarouris, line 473: here again if I change the nature of the input (to being a number) the test succeeds.

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -450,6 +472,62 @@ void handlerReturnsProjectsWrapperWithCorrectNextResultsAndPreviousResultsWhenLi
...
+
+        assertEquals(HttpURLConnection.HTTP_OK, gatewayResponse.getStatusCode());
+        assertEquals(MediaType.APPLICATION_JSON, gatewayResponse.getHeaders().get(HttpHeaders.CONTENT_TYPE));
+    }
+
+    @Test
+    void handlerReturnsProjectsFromTitleSearchWhenSuppliedWithQueryStringIncludingNumber() throws Exception {
+        cristinApiClientStub = spy(cristinApiClientStub);
+
+        doThrow(RuntimeException.class)
```
axthosarouris, line 467: I can change the input value to anything and the tests succeed any time. I can delete the first mocking statement and the tests are succeeding again.

I think the sequence of requests is not tested in the tests and I think this is what we are interested in. The exception that is thrown by the mock is not observed in any way. I would suggest to verify that the exception is thrown and that afterwards the response is received. Otherwise I can change the inputStream of the test and the test will succeed every time.

axthosarouris: The same holds for all tests

axthosarouris (changes requested):

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
torarnet: Must have made som errors writing those tests, i will fix

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
torarnet: See other test named: `handlerReturnsProjectsFromTitleSearchWhenGrantIdSearchReturnsZeroResults`

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
torarnet: See latest commit b8a752d

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
torarnet: It was a bug. It was to throw an exception if it does a query on title because it should only do query on grant id and not title. It should work now after latest commit

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
torarnet: It throws an exception if a query using grant id is called and if that happens the test should fail. So if we dont get an exception, it means it only queries using title which it should. It got updated in latest commit.

axthosarouris (approved):

brinxmat (approved):

PR #56, opened 2021-06-07
NP-2660: Title with whitespace now working

torarnet:
Replaces spaces with + symbol according to W3C spec


axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/UriUtils.java
```diff
@@ -14,6 +14,10 @@
     private static final String PARAMETER_DELIMITER = "&";
     private static final String EMPTY_QUERY_PARAMETERS_FOR_URI_CONSTRUCTOR = null;
 
+    public static String escapeWhiteSpace(String text) {
+        return text.replaceAll(" +", "+");
```
brinxmat, line 18: maybe "\s+" and extract constants to make it clear what is matched here and what the replacement is

brinxmat (changes requested):

janbj (approved):

axthosarouris (approved):

brinxmat (approved):

PR #60, opened 2021-06-15
NP-2280: Changed cfn template from hardcoded values to using parameters

torarnet:
Refactored code to comply with new changes in cfn template
Set default parameters for simplified testing and deployment

### DIFF
--- a/src/main/java/no/unit/nva/cristin/projects/Constants.java
+++ b/src/main/java/no/unit/nva/cristin/projects/Constants.java
@@ -10,22 +10,20 @@ public class Constants {
 
     public static final ObjectMapper OBJECT_MAPPER = JsonUtils.objectMapper;
     private static final Environment ENVIRONMENT = new Environment();
-    private static final String CRISTIN_API_HOST_ENV = "CRISTIN_API_HOST";
     public static final String PROJECT_SEARCH_CONTEXT_URL = "https://example.org/search-api-context.json";
     public static final String PROJECT_LOOKUP_CONTEXT_URL = "https://example.org/project-context.json";
-    public static final String CRISTIN_API_HOST = ENVIRONMENT.readEnv(CRISTIN_API_HOST_ENV);
+    public static final String CRISTIN_API_HOST = ENVIRONMENT.readEnvOpt("CRISTIN_API_HOST")
+        .orElse("api.cristin.no");
     public static final String CRISTIN_API_BASE_URL = "https://" + CRISTIN_API_HOST + "/v2";
     public static final String HTTPS = "https";
     public static final String EMPTY_FRAGMENT = null;
     public static final String INSTITUTION_PATH = "institutions";
     public static final String PERSON_PATH = "persons";
-    private static final String BASE_PATH_ENV = "BASE_PATH";
-    public static final String BASE_PATH = ENVIRONMENT.readEnv(BASE_PATH_ENV);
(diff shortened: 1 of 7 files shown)


brinxmat (approved):

src/main/java/no/unit/nva/cristin/projects/Constants.java
```diff
@@ -19,12 +19,10 @@
     public static final String EMPTY_FRAGMENT = null;
     public static final String INSTITUTION_PATH = "institutions";
     public static final String PERSON_PATH = "persons";
-    private static final String BASE_URL_ENV = "BASE_URL";
-    public static final String BASE_URL = ENVIRONMENT.readEnv(BASE_URL_ENV);
+    private static final String BASE_PATH_ENV = "BASE_PATH";
+    public static final String BASE_PATH = ENVIRONMENT.readEnv(BASE_PATH_ENV);
```
axthosarouris, line 23: I suggest that we assign a default value if the value is missing, because in case we do not setup the correct env variables. The reason is that  I am not sure whether things would show up in the logs when errors occur in the static initialization of the code. 
Also, I do not think it is necessary to declare a static variable in this case. Just use the string literal directly.

src/main/java/no/unit/nva/cristin/projects/UriUtils.java
```diff
@@ -24,6 +28,16 @@ public static URI buildUri(String... parts) {
         return attempt(() -> new URI(String.join(SLASH_DELIMITER, parts))).orElseThrow();
     }
 
+    public static URI getNvaProjectUriWithParams(Map<String, String> parameters) {
```
axthosarouris, line 31: is the delimiter at the end necessary?

axthosarouris (changes requested):

src/main/java/no/unit/nva/cristin/projects/UriUtils.java
torarnet: Will keep for now

axthosarouris (approved):

src/main/java/no/unit/nva/cristin/projects/Constants.java
janbj: Are we sure default valyes are a good thing, isn't it better to fail fast if something is misconfigured?

template.yaml
```diff
@@ -19,10 +19,9 @@ Globals:
     Environment:
       Variables:
         ALLOWED_ORIGIN: '*'
-        CRISTIN_API_HOST: 'api.cristin.no'
-        DOMAIN_NAME: 'api.dev.nva.aws.unit.no' # TODO: NP-2280: Dynamic from CustomDomain
-        # TODO: NP-2280: Replace with CustomDomainBasePath and call it PROJECTS_BASE_PATH
-        BASE_URL: 'https://api.dev.nva.aws.unit.no/project'
+        CRISTIN_API_HOST: !Ref CristinApiHost
+        DOMAIN_NAME: !Ref CustomDomain
```
janbj, line 23: Do we need separate DOMAIN_NAME and BASE_PATH? Can we do !Sub 'https://${CustomDomain}/${CustomDomainBasePath}/'?

torarnet: As far as I remember we discussed it and rather agreed on having them as separate. We should ask @brinxmat about it.

brinxmat (approved):

janbj (approved):

PR #63, opened 2021-06-21
Np 2315 karate tests add json ld

torarnet:


src/main/java/no/unit/nva/cristin/projects/CristinHandler.java
```diff
@@ -14,9 +17,19 @@
 
     protected static final String DEFAULT_LANGUAGE_CODE = "nb";
     private static final Set<String> VALID_LANGUAGE_CODES = Set.of("en", "nb", "nn");
+    private static final Set<String> VALID_CONTENT_TYPES = Set.of("application/json", "application/ld+json");
```
axthosarouris, line 20: can we add these types in commons?

src/main/java/no/unit/nva/cristin/projects/CristinHandler.java
```diff
@@ -30,4 +43,18 @@ protected static String getValidLanguage(RequestInfo requestInfo) throws BadRequ
...
+            .filter(VALID_CONTENT_TYPES::contains)
+            .ifPresent(this::addContentTypeHeader);
+    }
+
+    private void addContentTypeHeader(String contentType) {
+        this.additionalHeaders.put(HttpHeaders.CONTENT_TYPE, contentType);
+    }
+
+    private Optional<String> getRequestedContentType(RequestInfo requestInfo) {
+        return Optional.ofNullable(requestInfo.getHeaders().get(HttpHeaders.ACCEPT));
```
axthosarouris, line 58: Optional.ofNullable(requestInfo).map(r->r.getHeaders()).map(h->h.get(HttpHeaders.Accept)). Better to chain the maps in order to avoid null pointer exceptions.

axthosarouris: The '-> ' that I used was only for brevity. Use proper style and variable names.

src/main/java/no/unit/nva/cristin/projects/CristinHandler.java
torarnet: One of them already is but the other, ld json is missing. Should I commit to commons or somebody else? I have never worked with commons before

torarnet:
Needs more work. Specification says to throw exception if unsupported content type in 'Accept' header

src/main/java/no/unit/nva/cristin/projects/CristinHandler.java
axthosarouris: Add it to the commons

axthosarouris (approved):

axthosarouris:
The title of the PR does not make sense. There are no karate tests in the PR

axthosarouris (approved):

PR #65, opened 2021-06-22
NP-2315: Throw Bad request if unsupported query param

torarnet:


src/main/java/no/unit/nva/cristin/projects/FetchOneCristinProject.java
```diff
@@ -36,12 +41,20 @@ public FetchOneCristinProject(CristinApiClient cristinApiClient, Environment env
...
         String id = getValidId(requestInfo);
 
         return getTransformedProjectFromCristin(id, language);
     }
 
+    private void validateThatSuppliedQueryParamsIsSupported(RequestInfo requestInfo) throws BadRequestException {
+        if (!VALID_QUERY_PARAMS.containsAll(requestInfo.getQueryParameters().keySet())) {
+            throw new BadRequestException(ERROR_MESSAGE_INVALID_QUERY_PARAMS_ON_LOOKUP);
+        }
+    }
```
axthosarouris, line 56: This is the same code as the one above. Would it be worth to reuse the method?

axthosarouris (changes requested):

src/test/java/no/unit/nva/cristin/projects/FetchCristinProjectsTest.java
```diff
@@ -508,6 +511,22 @@ void handlerReturnsCristinProjectsWhenQueryContainsTitleWithWhitespace() throws
...
+            INVALID_QUERY_PARAM_KEY, INVALID_QUERY_PARAM_VALUE,
+            QUERY, RANDOM_TITLE));
+
+        handler.handleRequest(input, output, context);
+
+        GatewayResponse<Problem> gatewayResponse = GatewayResponse.fromOutputStream(output);
+
+        assertEquals(HttpURLConnection.HTTP_BAD_REQUEST, gatewayResponse.getStatusCode());
+        assertEquals(APPLICATION_PROBLEM_JSON, gatewayResponse.getHeaders().get(HttpHeaders.CONTENT_TYPE));
+        assertThat(gatewayResponse.getBodyObject(Problem.class).getDetail(),
```
axthosarouris, line 526: extract the body object into a variable

src/test/java/no/unit/nva/cristin/projects/FetchOneCristinProjectTest.java
```diff
@@ -212,6 +216,23 @@ void handlerThrowsInternalErrorWhenHttpStatusCodeIsSomeUnexpectedValue() throws
...
+            .withQueryParameters(Map.of(INVALID_QUERY_PARAM_KEY, INVALID_QUERY_PARAM_VALUE))
+            .withPathParameters(Map.of(ID, DEFAULT_ID))
+            .build();
+        handler.handleRequest(input, output, context);
+
+        GatewayResponse<Problem> gatewayResponse = GatewayResponse.fromOutputStream(output);
+
+        assertEquals(HttpURLConnection.HTTP_BAD_REQUEST, gatewayResponse.getStatusCode());
+        assertEquals(APPLICATION_PROBLEM_JSON, gatewayResponse.getHeaders().get(HttpHeaders.CONTENT_TYPE));
+        assertThat(gatewayResponse.getBodyObject(Problem.class).getDetail(),
```
axthosarouris, line 232: extact the body object as variable

axthosarouris (changes requested):

src/main/java/no/unit/nva/cristin/projects/FetchOneCristinProject.java
torarnet: There is 2 different Set of VALID_QUERY_PARAMS. If moved to CristinHandler I need to pass both Set and Error message as well as requestInfo to the method. Do you want me to do that or keep it as is?

axthosarouris: leave as is

axthosarouris (approved):

PR #83, opened 2021-11-05
Np 3380 model classes

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/PersonUriUtils.java
```diff
@@ -0,0 +1,33 @@
...
+            queryParameters(parameters), EMPTY_FRAGMENT)).orElseThrow();
+    }
+
+    public static URI getPersonUriWithId(String id) {
+        return attempt(() -> new URI(HTTPS, DOMAIN_NAME, getCommonPath() + id,
+            EMPTY_FRAGMENT)).orElseThrow();
+    }
+
+    private static String getCommonPath() {
+        return SLASH_DELIMITER + BASE_PATH + SLASH_DELIMITER + PERSON + SLASH_DELIMITER;
```
axthosarouris, line 31: Use UriWrapper instead, and possibly expand to add query parameter handling.

axthosarouris, line 31: Use UriWrapper instead, and possibly expand to add query parameter handling. I will come up with an example on Monday.

axthosarouris (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Role.java
```diff
@@ -0,0 +1,86 @@
...
+            this.labels = labels;
+            return this;
+        }
+
+        public Role build() {
+            return new Role(this.id, this.labels);
+        }
+    }
+
+    public static Role fromCristinAffiliation(CristinAffiliation cristinAffiliation) {
```
brinxmat, line 82: Move the logic to CristinAffiliation.toAffiliation

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
```diff
@@ -0,0 +1,209 @@
...
+            return new Person(this.context, this.id, this.identifiers, this.names, this.contactDetails,
+                this.image, this.affiliations);
+        }
+    }
+
+    public void setContext(String context) {
+        this.context = context;
+    }
+
+    private static ContactDetails extractContactDetails(CristinPerson cristinPerson) {
```
brinxmat, line 197: The methods for constructing this object should be part of CristinPerson and called in CristinPerson::toPerson

brinxmat (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
```diff
@@ -0,0 +1,209 @@
...
+     * @param context        The ontology context.
+     * @param id             Identifier of Person.
+     * @param identifiers    Different identifiers related to this object.
+     * @param names          Different names for this Person.
+     * @param contactDetails How to contact this Person.
+     * @param image          URI to picture of this Person.
+     * @param affiliations   This person's organization affiliations.
+     */
+    @JsonCreator
+    public Person(@JsonProperty("@context") String context, @JsonProperty("id") URI id,
```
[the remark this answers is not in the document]
torarnet: But that will change context for all instances of this class? What if we want the context field to be on the SearchResponse for a list of Person rather than on the Person itself?

cristin-person/src/main/java/no/unit/nva/cristin/person/PersonUriUtils.java
axthosarouris: Example method of expanding UriWrapper in commons:
```
    public UriWrapper addQueryParameter(String param, String value) throws URISyntaxException {
        var queryString = StringUtils.isBlank(uri.getQuery())
                              ? param + "=" + value
                              : "&" + param + "=" + value;
        URI newUri = new URI(uri.getScheme(),uri.getHost(),uri.getPath(),queryString,EMPTY_FRAGMENT);
        return new UriWrapper(newUri);
    }
```
URI does the decoding of the values itself

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinAffiliation.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Map;
+import nva.commons.core.JacocoGenerated;
+
+@SuppressWarnings("unused")
+@JacocoGenerated
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public class CristinAffiliation {
+
+    private Boolean active;
+    private Map<String, String> position;
```
axthosarouris, line 15: It is not clear what a position is. A class could be helpful.

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinUnit.java
```diff
@@ -0,0 +1,30 @@
+package no.unit.nva.cristin.person.model.cristin;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import nva.commons.core.JacocoGenerated;
+
+@SuppressWarnings("unused")
+@JacocoGenerated
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
```
axthosarouris, line 9: Why do we need a special naming strategy?

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Affiliation.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Objects;
+import no.unit.nva.cristin.person.model.cristin.CristinAffiliation;
+import nva.commons.core.JacocoGenerated;
+
+@JacocoGenerated
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class Affiliation {
+
+    private final URI organization;
+    private final Boolean active;
```
axthosarouris, line 17: Boolean can be null, what is the meaning of null?

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Affiliation.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @JacocoGenerated
+    @Override
+    public int hashCode() {
+        return Objects.hash(getOrganization(), getActive(), getRole());
+    }
+
+    @JacocoGenerated
+    public static final class Builder {
```
axthosarouris, line 69: Is Builder necessary for an object of 3 values? In any case a it is better to have Builders that mutate the target object using the object's setters. So each withX() calls setX() and build() just returns the object.

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/ContactDetails.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @JacocoGenerated
+    @Override
+    public int hashCode() {
+        return Objects.hash(getTelephone());
+    }
+
+    @JacocoGenerated
+    public static final class Builder {
```
axthosarouris, line 42: Builder is an overkill for an object with only one property

axthosarouris (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/NvaIdentifier.java
```diff
@@ -0,0 +1,125 @@
...
+
+        Optional.of(cristinPerson)
+            .map(CristinPerson::getFirstNamePreferred)
+            .ifPresent(id -> names.add(createOneIdentifier("PreferredFirstName", id)));
+
+        Optional.of(cristinPerson)
+            .map(CristinPerson::getSurnamePreferred)
+            .ifPresent(id -> names.add(createOneIdentifier("PreferredLastName", id)));
+
+        return names;
```
axthosarouris, line 119: It seems that these are the names of the fields of the CirstinPerson. This code belongs to CristinPerson, basically I agree with Rurik that the "fromX" methods should be "toX" methods. But in any case it is better to use constants instead of magic strings.

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinAffiliation.java
torarnet: But the problem is that we need dynamic field names (different language codes) and when using a map we can dynamically set these using keys

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinUnit.java
torarnet: Data from Cristin use underscore case

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Affiliation.java
torarnet: You want me to use primitive instead?

cristin-person/src/main/java/no/unit/nva/cristin/person/PersonUriUtils.java
torarnet: You want me to make the change in nva-commons library?

cristin-person/src/test/java/no/unit/nva/cristin/person/model/cristin/CristinPersonTest.java
```diff
@@ -0,0 +1,71 @@
...
+public class CristinPersonTest {
+
+    private static final ObjectMapper OBJECT_MAPPER = JsonUtils.dtoObjectMapper;
+    private static final String CRISTIN_GET_PERSON_JSON =
+        "cristinGetPersonResponse.json";
+    private static final String NVA_API_GET_PERSON_JSON =
+        "nvaApiGetPersonResponse.json";
+
+    @Test
+    void cristinModelBuildsCorrectlyWhenDeserializingPersonJson() throws IOException {
```
axthosarouris, line 23:
```
  @Test
    void cristinModelBuildsCorrectlyWhenDeserializingPersonJson() throws IOException {
        String body = getBodyFromResource(CRISTIN_GET_PERSON_JSON);
        CristinPerson cristinPerson = fromJson(body, CristinPerson.class);
        JsonNode node = OBJECT_MAPPER.readTree(body);
        ObjectNode j1 = removeIgnoredFields(body);
        String s2 = OBJECT_MAPPER.writeValueAsString(cristinPerson);
        JsonNode j2 = OBJECT_MAPPER.readTree(s2);
        assertEquals(j1,j2);
    }

    private ObjectNode removeIgnoredFields(String body) throws JsonProcessingException {
        ObjectNode j1 = (ObjectNode) OBJECT_MAPPER.readTree(body);
        j1.remove("identified_cristin_person");
        j1.remove("cristin_profile_url");
        removeInstiutionFieldFromAffiliations(j1);
        return j1;
    }

    private void removeInstiutionFieldFromAffiliations(ObjectNode j1) {
        ArrayNode affiliations= (ArrayNode) j1.get("affiliations");
        affiliations.forEach(af->removeInstitution(af));
    }

    private void removeInstitution(JsonNode af) {
        var objectNode =(ObjectNode)af;
        objectNode.remove("institution");
    }
```
I think it gives better overview of what you are trying to do

cristin-person/src/test/java/no/unit/nva/cristin/person/model/cristin/CristinPersonTest.java
```diff
@@ -0,0 +1,71 @@
...
+        String body = getBodyFromResource(CRISTIN_GET_PERSON_JSON);
+        CristinPerson cristinPerson = fromJson(body, CristinPerson.class);
+        JsonNode node = OBJECT_MAPPER.readTree(body);
+
+        assertEquals(cristinPerson.getCristinPersonId(), node.get("cristin_person_id").asText());
```
axthosarouris, line 28: assertEquals(expected,actual). Although the order does not matter when the test succeeds, it matters when it fails. This is why I usually prefer the more intuitive assertThat(actual,is(equalTo(expected)))

axthosarouris (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinAffiliation.java
axthosarouris: Still not clear to me what is a position

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPerson.java
```diff
@@ -0,0 +1,192 @@
...
+        List<TypedValue> names = new ArrayList<>();
+
+        Optional.ofNullable(getFirstName())
+            .ifPresent(id -> names.add(new TypedValue(FIRST_NAME, id)));
+        Optional.ofNullable(getSurname())
+            .ifPresent(id -> names.add(new TypedValue(LAST_NAME, id)));
+        Optional.ofNullable(getFirstNamePreferred())
+            .ifPresent(id -> names.add(new TypedValue(PREFERRED_FIRST_NAME, id)));
+        Optional.ofNullable(getSurnamePreferred())
+            .ifPresent(id -> names.add(new TypedValue(PREFERRED_LAST_NAME, id)));
```
axthosarouris, line 179: probably it is the getters that should return Optional

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Affiliation.java
axthosarouris: yes

axthosarouris (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
```diff
@@ -0,0 +1,200 @@
...
+
+    public ContactDetails getContactDetails() {
+        return contactDetails;
+    }
+
+    public URI getImage() {
+        return image;
+    }
+
+    public void setContext(String thatContext) {
```
brinxmat, line 88: Is context not static?

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
```diff
@@ -0,0 +1,200 @@
...
+    @JacocoGenerated
+    public static final class Builder {
+
+        private final transient Person person;
+
+        public Builder() {
+            person = new Person();
+        }
+
+        public Builder withContext(String context) {
```
brinxmat, line 160: Static context again

cristin-person/src/test/java/no/unit/nva/cristin/person/model/cristin/CristinPersonTest.java
```diff
@@ -0,0 +1,76 @@
...
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.IOException;
+import java.nio.file.Path;
+import no.unit.nva.cristin.person.model.nva.Person;
+import nva.commons.core.JsonUtils;
+import nva.commons.core.ioutils.IoUtils;
+import org.junit.jupiter.api.Test;
+
+public class CristinPersonTest {
```
brinxmat, line 18: Generally, use the "shouldX" formulation for test names…it is clear what is tested. I assume these tests will be deleted anyway once the relevant handler is in place, but it is good practice to use the common naming format.

brinxmat (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
torarnet: Yes but i want the option to exclude it if we use it in a SearchResponse

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinAffiliation.java
torarnet: We discussed it and concluded we do not have a better way of creating dynamic field names for the moment, so now just renaming to a more descriptive value

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
torarnet: Do you have any suggestions to how i should make context optional? I want to use it in lookup but not in query.

brinxmat (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.stream.Collectors;
+import nva.commons.core.JacocoGenerated;
+
+@JacocoGenerated
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonPropertyOrder({"@context"})
+public class Person {
+
+    @JsonInclude(NON_NULL)
+    @JsonProperty("@context")
```
axthosarouris, line 24: Minor comment: in case of static fields, I would suggest putting the `@JsonProperty` at the getter and the setter making it empty. We avoid thus to have a field that is only a copy of a static value. It is also more clear about our intention.

axthosarouris (approved):

PR #86, opened 2021-11-15
Np 3392 person lookup handler dummy response

torarnet:


axthosarouris (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
```diff
@@ -76,8 +76,8 @@ public String getContext() {
         return context;
     }
 
-    public static void setContext(String context) {
-        Person.context = context;
+    public void setContext(String context) {
```
torarnet, line 79: Tests failed when using static because of the difference between a query and a lookup. One person context should not have a value in a query response since the query wrapper has its own context. When doing a build with both query and lookup it failed because of it

PR #87, opened 2021-11-18
Np 3419 replace lists with sets

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPerson.java
```diff
@@ -161,12 +162,12 @@ private URI extractIdUri() {
...
-     * Transforms names from a Cristin person into a List.
+     * Transforms names from a Cristin person into a Set.
      *
-     * @return List of transformed names
+     * @return Set of transformed names
      */
-    private List<TypedValue> extractNames() {
-        List<TypedValue> names = new ArrayList<>();
+    private Set<TypedValue> extractNames() {
+        Set<TypedValue> names = new HashSet<>();
```
brinxmat, line 170: Probably just "Set::of"

brinxmat (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPerson.java
torarnet: Could not find a good way to chain Optional::ifPresent into a Set::of. Suggestions?

`
        Set<TypedValue> names = new HashSet<>();

        names.add(new TypedValue(FIRST_NAME, getFirstName()));
        names.add(new TypedValue(LAST_NAME, getSurname()));
        getFirstNamePreferred().ifPresent(name -> names.add(new TypedValue(PREFERRED_FIRST_NAME, name)));
        getSurnamePreferred().ifPresent(name -> names.add(new TypedValue(PREFERRED_LAST_NAME, name)));

        return names;`

brinxmat (approved):

PR #94, opened 2021-11-29
Np 3360 query organization

svenngjeto:


cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +52,65 @@ public Organization getOrganization(URI uri)
...
+        Map<String, String> cristinRequestQueryParams = translateToCristinApi(requestQueryParams);
+        URI queryUri = new UriWrapper(HTTPS, CRISTIN_API_BASE)
+                .addChild(UNITS_PATH)
+                .addQueryParameters(cristinRequestQueryParams)
+                .getUri();
+
+        long start = System.currentTimeMillis();
+        SearchResponse<Organization> searchResponse = httpExecutor.query(queryUri);
+
+        return updateSearchResponseMetadata(searchResponse,requestQueryParams, System.currentTimeMillis() - start);
```
axthosarouris, line 66: create a variable totalProcessingTime

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +52,65 @@ public Organization getOrganization(URI uri)
...
+                PAGE, requestQueryParams.get(PAGE),
+                CRISTIN_PER_PAGE_PARAM, requestQueryParams.get(NUMBER_OF_RESULTS));
     }
+
+    private SearchResponse<Organization> updateSearchResponseMetadata(
+            SearchResponse<Organization> searchResponse,
+            Map<String, String> requestQueryParams,
+            long timeUsed) {
+        final URI baseUri = new UriWrapper(HTTPS, DOMAIN_NAME).addChild(BASE_PATH).addChild(ORGANIZATION_PATH).getUri();
+        searchResponse.setId(new UriWrapper(baseUri).addQueryParameters(requestQueryParams).getUri());
```
axthosarouris, line 81: refactor this command to a method `new UriWrapper(baseUri).addQueryParameters(requestQueryParams).getUri()`

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +52,65 @@ public Organization getOrganization(URI uri)
...
+            nextMap.put(PAGE, Integer.toString(firstPage));
+            return new UriWrapper(baseUri).addQueryParameters(nextMap).getUri();
+        }
+        return null;
+    }
+
+    private URI nextResult(URI baseUri, Map<String, String> requestQueryParams, int totalSize) {
+        int nextPage = Integer.parseInt(requestQueryParams.get(PAGE)) + 1;
+        int pageSize = Integer.parseInt(requestQueryParams.get(NUMBER_OF_RESULTS));
+        if (nextPage * pageSize < totalSize) {
```
axthosarouris, line 102: This line works only if the total number of results is an exact multiple of pageSize. if totalSize == 29 and pageSize==10 then nextPage*pageSize == 30 and the last 9 results will not be visible.

axthosarouris (changes requested):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +52,65 @@ public Organization getOrganization(URI uri)
...
+
+    private SearchResponse<Organization> updateSearchResponseMetadata(
+            SearchResponse<Organization> searchResponse,
+            Map<String, String> requestQueryParams,
+            long timeUsed) {
+        final URI baseUri = new UriWrapper(HTTPS, DOMAIN_NAME).addChild(BASE_PATH).addChild(ORGANIZATION_PATH).getUri();
+        searchResponse.setId(new UriWrapper(baseUri).addQueryParameters(requestQueryParams).getUri());
+        searchResponse.setFirstRecord(calculateFirstRecord(requestQueryParams));
+        searchResponse.setNextResults(nextResult(baseUri, requestQueryParams, searchResponse.getSize()));
+        searchResponse.setPreviousResults(previousResult(baseUri, requestQueryParams, searchResponse.getSize()));
```
torarnet, line 84: Can you use `usingHeadersAndQueryParams` in `SearchResponse` or do you have to create these some other way?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
torarnet: There is logic for this in `SearchResponse` but it depends on headers from Cristin response

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -32,6 +36,7 @@
 
 public class HttpExecutorImpl {
 
+    public static final ObjectMapper OBJECTMAPPER = JsonUtils.dtoObjectMapper;
```
torarnet, line 39: Isn't this a constant in commons?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -96,20 +100,41 @@ protected SubSubUnitDto fetch(URI uri) throws InterruptedException, NotFoundExce
...
+    protected SearchResponse<Organization> query(URI uri) throws NotFoundException, FailedHttpRequestException {
+        HttpRequest httpRequest = createHttpRequest(addLanguage(uri));
+        HttpResponse<String> response = sendRequest(httpRequest);
+        if (isSuccessful(response.statusCode())) {
+            try {
+                List<Organization> organizations = getOrganizations(response);
+                final int count =  response.headers().firstValue("X-Total-Count").isPresent()
+                        ? Integer.parseInt(response.headers().firstValue("X-Total-Count").get())
+                        : organizations.size();
+                return new SearchResponse<Organization>(uri).withHits(organizations).withSize(count);
```
torarnet, line 112: Should `uri` be Cristin uri or NVA uri?

torarnet (commented):
See comments

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
svenngjeto: Uri is overwritten with .setId(...) later... Id is against NVA API

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
svenngjeto: Indeed!

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
svenngjeto: Is rewriting uri/id to reference NVA API, not Cristin api

torarnet: What if you can get access to Nva Id when creating the searchResponse object, then it wont be necessary to create the logic one more time. Up to you

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
svenngjeto: Recalculating nextResult in another/better way...   At this point we cannot use the cristin headers as we rewrite the uri

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
svenngjeto: It is possible to calculate the id beforhand, but it is not in the header - but we could move the logic to searchResponse later

cristin-organization/src/test/java/no/unit/nva/cristin/organization/organization.feature
```diff
@@ -57,4 +57,4 @@ Feature: API tests for Cristin Organization retrieve and search
     Then status 200
     And match response.hits == '#array'
     And match response.size == '#number'
-    And match response.hits == '#[0]' // hits array length == 0
+    And match response.hits == '#[2]' // hits array length == 0
```
torarnet, line 60: 2? Outdated comment?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +52,74 @@ public Organization getOrganization(URI uri)
...
+        int pageSize = Integer.parseInt(requestQueryParams.get(NUMBER_OF_RESULTS));
+        if (currentPage * pageSize < totalSize) {
+            Map<String, String> nextMap = new ConcurrentHashMap<>(requestQueryParams);
+            nextMap.put(PAGE, Integer.toString(currentPage + 1));
+            return new UriWrapper(baseUri).addQueryParameters(nextMap).getUri();
+        }
+        return null;
+    }
+
+    private Integer calculateFirstRecord(Map<String, String> requestQueryParams) {
```
torarnet, line 119: This is now in both this class and SearchResponse class. Maybe we should reuse the one, `indexOfFirstEntryInPageCalculatedFromParams` in SearchResponse by making it public?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +52,74 @@ public Organization getOrganization(URI uri)
...
+        searchResponse.setFirstRecord(calculateFirstRecord(requestQueryParams));
+        searchResponse.setNextResults(nextResult(getNvaApiBaseUri(), requestQueryParams, searchResponse.getSize()));
+        searchResponse.setPreviousResults(previousResult(getNvaApiBaseUri(),
+                requestQueryParams,
+                searchResponse.getSize()));
+        searchResponse.setProcessingTime(timeUsed);
+        return searchResponse;
+    }
+
+    private URI getNvaQueryId(Map<String, String> requestQueryParams) {
```
torarnet, line 90: Can you reuse `createIdUriFromParams` in `UriUtils` ?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +52,74 @@ public Organization getOrganization(URI uri)
...
-                        .addChild(UriUtils.INSTITUTION)
-                        .addQueryParameters(requestQueryParams)
-                        .getUri())
-                .withProcessingTime(0L)
-                .withSize(0)
-                .withHits(Collections.emptyList());
+    public SearchResponse<Organization> queryOrganizations(Map<String, String> requestQueryParams)
+            throws NotFoundException, FailedHttpRequestException {
+        Map<String, String> cristinRequestQueryParams = translateToCristinApi(requestQueryParams);
+        URI queryUri = new UriWrapper(HTTPS, CRISTIN_API_BASE)
```
torarnet, line 58: Extract into reusable method in `UriUtils` ?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +52,74 @@ public Organization getOrganization(URI uri)
...
+
+    private URI getNvaQueryId(Map<String, String> requestQueryParams) {
+        return new UriWrapper(getNvaApiBaseUri()).addQueryParameters(requestQueryParams).getUri();
+    }
+
+    private URI getNvaApiBaseUri() {
+        return new UriWrapper(HTTPS, DOMAIN_NAME).addChild(BASE_PATH).addChild(ORGANIZATION_PATH).getUri();
+    }
+
+    private URI previousResult(URI baseUri, Map<String, String> requestQueryParams, int totalSize) {
```
torarnet, line 98: Skeptical about duplicating the code when there is something similar in SearchResponse, but that can be refactored later.

torarnet (commented):
See comments

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
svenngjeto: Of course...

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
torarnet: Or the other way around, reuse you method in SearchResponse, but then it have to be moved to some common class

torarnet: I think since this logic is used in SearchResponse the logic belongs to that class. For example a static functional method in SearchResponse which can be used everywhere and does not rely on state

torarnet: What if there are zero results? `(1 - 1) * 5 + 1 = 1`

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -96,20 +98,41 @@ protected SubSubUnitDto fetch(URI uri) throws InterruptedException, NotFoundExce
         }
     }
 
+    protected SearchResponse<Organization> query(URI uri) throws NotFoundException, FailedHttpRequestException {
+        HttpRequest httpRequest = createHttpRequest(addLanguage(uri));
+        HttpResponse<String> response = sendRequest(httpRequest);
+        if (isSuccessful(response.statusCode())) {
+            try {
+                List<Organization> organizations = getOrganizations(response);
+                final int count =  response.headers().firstValue("X-Total-Count").isPresent()
```
brinxmat, line 107: Extract

brinxmat (changes requested):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
svenngjeto: It creates uri against cristin api, not nva - we have to make it neutral to the uri

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
svenngjeto: Checking if we have hits

brinxmat (approved):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinApiClient.java
```diff
@@ -45,15 +53,68 @@ public Organization getOrganization(URI uri)
...
+                searchResponse.getSize()));
+        searchResponse.setProcessingTime(timeUsed);
+        return searchResponse;
+    }
+
+    private URI getNvaApiBaseUri() {
+        return new UriWrapper(HTTPS, DOMAIN_NAME).addChild(BASE_PATH).addChild(ORGANIZATION_PATH).getUri();
     }
+
+    private URI previousResult(URI baseUri, Map<String, String> requestQueryParams, int totalSize) {
```
torarnet, line 93: Maybe we can put these in SearchResponse later to be reused

cristin-organization/src/test/java/no/unit/nva/cristin/organization/QueryCristinOrganizationHandlerTest.java
```diff
@@ -97,25 +107,48 @@ void shouldReturnEmptyResponseOnStrangeQuery() throws IOException {
...
-        cristinApiClient = new CristinApiClient();
+        HttpExecutorImpl httpExecutor = new HttpExecutorImpl(httpClient);
+        output = new ByteArrayOutputStream();
+
+        HttpExecutorImpl mySpy = spy(httpExecutor);
+        final URI first = getCristinUri("185.53.18.14", UNITS_PATH);
+        doReturn(getOrganization("org_18_53_18_14.json")).when(mySpy).getOrganization(first);
+        final URI second = getCristinUri("2012.9.20.0", UNITS_PATH);
+        doReturn(getOrganization("org_2012_9_20_0.json")).when(mySpy).getOrganization(second);
+        HttpResponse<String> mockHttpResponse = mock(HttpResponse.class);
```
torarnet, line 125: There is also a HttpResponseFaker in cristin-commons that can be used

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -96,20 +98,47 @@ protected SubSubUnitDto fetch(URI uri) throws InterruptedException, NotFoundExce
...
 
-    private HttpResponse<String> sendRequest(HttpRequest httpRequest) throws InterruptedException,
-            FailedHttpRequestException {
-        try {
-            return httpClient.sendAsync(httpRequest, BodyHandlers.ofString()).get();
-        } catch (ExecutionException e) {
-            throw new FailedHttpRequestException(e);
-        }
+    protected HttpResponse<String> sendRequest(HttpRequest httpRequest) {
+        return attempt(() -> httpClient.sendAsync(httpRequest, BodyHandlers.ofString()).get()).orElseThrow();
```
torarnet, line 141: Not sure if you could use some parts from ApiClient in cristin-commons in this class

torarnet (approved):
Approved with minor comments

PR #96, opened 2021-11-29
Np 2749 search by person name

torarnet:


cristin-person/src/test/java/no/unit/nva/cristin/person/handler/PersonQueryHandlerTest.java
```diff
@@ -58,6 +82,50 @@ void shouldReturnResponseWhenCallingEndpointWithNameParameter() throws IOExcepti
         assertEquals(expected, actual);
     }
 
+    @Test
+    void shouldThrowCorrectErrorResponseToClientWhenBackendFetchFails() throws IOException {
```
axthosarouris, line 86: I am not sure this is as broad as it claims to be. It only tests the case where a RuntimeException produces a Runtime error. If there are more cases I suggest writing a test for each case. If there aren't more cases then rename the test

axthosarouris (changes requested):

brinxmat (commented):
See Orestis' comment

cristin-person/src/test/java/no/unit/nva/cristin/person/handler/PersonQueryHandlerTest.java
torarnet: Updated name and added extra test.

brinxmat (dismissed):

svenngjeto (dismissed):

.github/workflows/karate.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Run Karate tests erikbos/karate-action@main
         uses: erikbos/karate-action@main
         with:
-          url: https://api.dev.nva.aws.unit.no/cristin-karate/
+          url: https://api.dev.nva.aws.unit.no/np2749-karate-cristin/
```
axthosarouris, line 29: Does this mean that every time we need to run the karate tests we have to update the uri?

svenngjeto (approved):

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/HttpResponseFaker.java
```diff
@@ -19,54 +19,54 @@
 import static no.unit.nva.cristin.model.Constants.REL_PREV;
 import static no.unit.nva.cristin.model.Constants.X_TOTAL_COUNT;
 
-
 @JacocoGenerated
-public class HttpResponseStub implements HttpResponse<String> {
+public class HttpResponseFaker implements HttpResponse<String> {
```
axthosarouris, line 23: I noticed this is class is in main folder. If we wish to share testing logic among projects, we should create a "common-testing" module where we import it as test dependency. We do not mix production and test code.

cristin-organization/src/test/java/no/unit/nva/cristin/organization/organization.feature
```diff
@@ -3,7 +3,7 @@ Feature: API tests for Cristin Organization retrieve and search
   Background:
     * def SERVER_URL = 'https://api.dev.nva.aws.unit.no'
     * def testOrganizationNameSearchTerm = 'univers'
-    * def CRISTIN_BASE = SERVER_URL + '/cristin-karate'
+    * def CRISTIN_BASE = SERVER_URL + '/np2749-karate-cristin'
```
axthosarouris, line 6: Add a comment why this is changing

samconfig.toml
```diff
@@ -2,11 +2,11 @@ version = 0.1
...
-stack_name = "karate-cristin-service"
+stack_name = "np2749-karate-cristin-service"
 s3_bucket = "aws-sam-cli-managed-default-samclisourcebucket-plnc75nq3gkx"
-s3_prefix = "karate-cristin-service-np3360"
+s3_prefix = "np2749-karate-cristin-service"
 region = "eu-west-1"
 confirm_changeset = false
 capabilities = "CAPABILITY_IAM"
-parameter_overrides = "CustomDomain=\"api.dev.nva.aws.unit.no\" CustomDomainBasePath=\"cristin-karate\" CristinApiHost=\"api.cristin.no\""
+parameter_overrides = "ApiDomain=\"/NVA/ApiDomain\" CustomDomainBasePath=\"np2749-karate-cristin\" CristinApiHost=\"api.cristin.no\""
```
axthosarouris, line 11: you do not need override by writing again the default value.

axthosarouris (changes requested):

.github/workflows/karate.yml
torarnet: @svenngjeto is working on it, but for the time being we have to do it this way.

samconfig.toml
torarnet: This is autogenerated

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/HttpResponseFaker.java
torarnet: Moved to cristin-testing

axthosarouris (approved):

brinxmat (approved):

PR #98, opened 2021-12-03
Np 2820 get person by identifier

torarnet:


cristin-commons/src/main/java/no/unit/nva/cristin/common/handler/CristinQueryHandler.java
```diff
@@ -21,7 +21,7 @@
 
     private static final Set<String> VALID_QUERY_PARAMS = Set.of(QUERY, PAGE, NUMBER_OF_RESULTS);
 
-    public static final String WORD_WHITESPACE_DASH_COMMA_PERIOD = "[\\w\\s\\-,.]+";
+    public static final String WORD_WHITESPACE_DASH_COMMA_PERIOD = "[\\w\\s\\-,.æøå]+";
```
brinxmat, line 24: What do we define as a "valid query"? 

If the intention is to avoid something (like SQL injection?), then we need perhaps to look at this again.

Alternatively, any letter in any letter-based character set (includes Latin, Greek, Russian and probably others)
`^[\p{L}\x20\-,.]+$`

On the other hand, if we want any Latin letter, it becomes:

`^[A-Za-z\\u00C0-\\u00D6\\u00D8-\\u00f6\\u00f8-\\u00ff\x20\-,.]+$`

I guess it is dependent on what your match is trying to achieve and what the content is allowed to be. If it's names, then some names may. have e.g. 'ç, þ', etc.

cristin-person/src/main/java/no/unit/nva/cristin/person/handler/PersonFetchHandler.java
```diff
@@ -50,14 +50,15 @@ protected Integer getSuccessStatusCode(Void input, Person output) {
...
         }
     }
 
     private String getValidId(RequestInfo requestInfo) throws BadRequestException {
-        attempt(() -> Integer.parseInt(requestInfo.getPathParameter(ID)))
-            .orElseThrow(failure -> new BadRequestException(ERROR_MESSAGE_INVALID_PATH_PARAMETER_FOR_ID));
-
-        return requestInfo.getPathParameter(ID);
+        String identifier = requestInfo.getPathParameter(ID);
+        if (!Utils.isPositiveInteger(identifier)) {
```
brinxmat, line 59: Here, the usage is the exact opposite of everywhere else in the project. I don't disagree that throwing the exception should be the exception, though. I'd refactor the method to Utils:isNegativeInteger and the rest of the code in the project to match this pattern in another PR.

cristin-person/src/test/java/no/unit/nva/cristin/person/client/CristinPersonQueryTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static no.unit.nva.cristin.model.JsonPropertyNames.NUMBER_OF_RESULTS;
+import static no.unit.nva.cristin.model.JsonPropertyNames.PAGE;
+import static no.unit.nva.cristin.model.JsonPropertyNames.QUERY;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import java.net.URI;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class CristinPersonQueryTest {
```
brinxmat, line 11: This test is tests things that are tested elsewhere, I would remove it (and any other non-handler tests), or create a test in the handler that exposes what you're trying to test here.

brinxmat (changes requested):

brinxmat (commented):
PersonFetchHandler -> FetchPersonHandler, which matches the pattern used in the rest of the code

cristin-person/src/test/java/no/unit/nva/cristin/person/client/CristinPersonQueryTest.java
torarnet: It is not easy to test in another way  because it is only used internally and not returned to the client

axthosarouris (commented):
No further comments than those already pointed out

cristin-commons/src/main/java/no/unit/nva/cristin/common/handler/CristinQueryHandler.java
torarnet: Having problems with numbers (for grant id search) and æøå with these regex. Maybe we should do as discussed and only check not blank?

cristin-person/src/test/java/no/unit/nva/cristin/person/client/CristinPersonQueryTest.java
torarnet: Found a way to test. Deleted class

axthosarouris (approved):

brinxmat (approved):

PR #102, opened 2021-12-08
Np 2819 fetch person from orcid

torarnet:
A little skeptical about how to set URI to cristin test instance, as orcid fetch is not in cristin production yet

### DIFF
--- a/cristin-commons/src/main/java/no/unit/nva/cristin/model/Constants.java
+++ b/cristin-commons/src/main/java/no/unit/nva/cristin/model/Constants.java
@@ -48,7 +48,7 @@ public class Constants {
     public static final String NOT_FOUND_MESSAGE_TEMPLATE = "The URI \"%s\" cannot be dereferenced";
     public static final String QUERY_PARAMETER_LANGUAGE = "lang";
     public static final String ALL_QUERY_PARAMETER_LANGUAGES = "en,nb,nn";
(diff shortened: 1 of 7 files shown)


cristin-commons/src/main/java/no/unit/nva/cristin/model/Constants.java
```diff
@@ -44,6 +48,7 @@
     public static final String NOT_FOUND_MESSAGE_TEMPLATE = "The URI \"%s\" cannot be dereferenced";
     public static final String QUERY_PARAMETER_LANGUAGE = "lang";
     public static final String ALL_QUERY_PARAMETER_LANGUAGES = "en,nb,nn";
+    public static final Pattern ORCID_PATTERN = Pattern.compile("[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{4}");
```
brinxmat, line 51: I think that the ORCID pattern should be validated properly as "Both ORCID and ISNI use 16-character identifiers, using the digits 0–9, and separated into groups of four by hyphens. The final character, which may also be a letter "X" representing the value "10" (for example, Stephen Hawking's ORCID iD is https://orcid.org/0000-0002-9079-593X), is a MOD 11-2 check digit conforming to the ISO/IEC 7064:2003 standard."

brinxmat (changes requested):

axthosarouris (approved):

brinxmat (approved):

PR #109, opened 2021-12-16
Np 3491 more org haspart data

svenngjeto:


.github/workflows/karate.yml
```diff
@@ -28,9 +28,12 @@ jobs:
       # Prevent prompts and failure when the stack is unchanged
       - name: Deploy
         run: sam deploy --no-confirm-changeset --no-fail-on-empty-changeset  --stack-name ${{env.STACK_NAME}} --parameter-overrides ParameterKey=CustomDomain,ParameterValue=${DOMAIN_NAME} ParameterKey=CustomDomainBasePath,ParameterValue=${{env.BASE_PATH}} ParameterKey=CristinApiHost,ParameterValue=${CRISTIN_API}
+      # Take a break wainting for API to deploy
+      - name: Sleep for 30 seconds
+        run: sleep 30s
```
torarnet, line 33: Nice

.github/workflows/karate.yml
```diff
@@ -28,9 +28,12 @@ jobs:
...
       # Run karate tests on deployed api
       - name: Run karate tests with Gradle
         run: BASE_PATH=${{env.BASE_PATH}} ./gradlew karateTest -x test --info
       # Delete stack after running tests
-      - name: Delete stack after running tests
-        run: sam delete --no-prompts --region eu-west-1 --stack-name ${{env.STACK_NAME}}
+#      - name: Delete stack after running tests
+#        run: sam delete --no-prompts --region eu-west-1 --stack-name ${{env.STACK_NAME}}
```
torarnet, line 39: Uncomment before merge

.github/workflows/karate.yml
```diff
@@ -28,9 +28,12 @@ jobs:
       # Prevent prompts and failure when the stack is unchanged
       - name: Deploy
         run: sam deploy --no-confirm-changeset --no-fail-on-empty-changeset  --stack-name ${{env.STACK_NAME}} --parameter-overrides ParameterKey=CustomDomain,ParameterValue=${DOMAIN_NAME} ParameterKey=CustomDomainBasePath,ParameterValue=${{env.BASE_PATH}} ParameterKey=CristinApiHost,ParameterValue=${CRISTIN_API}
+      # Take a break wainting for API to deploy
```
torarnet, line 31: waiting*

cristin-commons/src/main/java/no/unit/nva/cristin/model/Constants.java
```diff
@@ -45,7 +45,7 @@
     public static final String PERSON_PATH_NVA = "person";
     public static final String PERSON_QUERY_CONTEXT = "https://example.org/person-search-context.json";
     public static final String UNITS_PATH = "units";
-    public static final String NOT_FOUND_MESSAGE_TEMPLATE = "The URI \"%s\" cannot be dereferenced";
+    public static final String NOT_FOUND_MESSAGE_TEMPLATE = "The URI '%s' cannot be dereferenced";
```
brinxmat, line 48: URI -> resource

cristin-organization/src/test/java/no/unit/nva/cristin/organization/QueryCristinOrganizationHandlerTest.java
```diff
@@ -147,6 +148,10 @@ void shouldReturnResponseOnQuery() throws IOException, NotFoundException, Failed
         assertEquals(JSON_UTF_8.toString(), gatewayResponse.getHeaders().get(HttpHeaders.CONTENT_TYPE));
     }
 
+    private Try<HttpResponse<String>> getaTry(HttpResponse<String> mockHttpResponse) {
```
brinxmat, line 151: getAttempt? or getTry

cristin-organization/src/test/java/no/unit/nva/cristin/organization/organization.feature
```diff
@@ -47,7 +47,7 @@ Feature: API tests for Cristin Organization retrieve and search
     Then status 404
     And match response.title == 'Not Found'
     And match response.status == 404
-    And match response.detail == 'The URI "' + CRISTIN_BASE + '/organization/0.1.2.3" cannot be dereferenced'
+    And match response.detail == 'The URI \'' + CRISTIN_BASE + '/organization/0.1.2.3\' cannot be dereferenced'
```
brinxmat, line 50: URI -> resource

brinxmat (changes requested):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -67,83 +118,135 @@ private Organization fromSubSubunit(SubSubUnitDto subSubUnitDto) {
...
+
+        subsubUnitDto.setSourceUri(subunitUri);
+        return subsubUnitDto;
     }
 
+    protected HttpResponse<String> throwExceptionIfNotSuccessful(HttpResponse<String> response, URI requestedUri)
+            throws FailedHttpRequestException, NotFoundException {
+        if (isNull(response)) {
+            throw new FailedHttpRequestException(NULL_HTTP_RESPONSE_ERROR_MESSAGE);
+        } else if (response.statusCode() >= FIRST_SUCCESSFUL_CODE
```
torarnet, line 238: Reuse isSuccessful?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -67,83 +118,135 @@ private Organization fromSubSubunit(SubSubUnitDto subSubUnitDto) {
...
+                && response.statusCode() < FIRST_NON_SUCCESSFUL_CODE) {
+            return response;
+        } else if (response.statusCode() == HTTP_NOT_FOUND) {
+            throw new NotFoundException(String.format(NOT_FOUND_MESSAGE_TEMPLATE, requestedUri));
+        } else {
+            throw new FailedHttpRequestException(errorMessage(response));
```
torarnet, line 244: Error messages is used internally right? I can see you throw a client ApiGatewayException in handleError so I assume this is internal

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -52,11 +69,45 @@ public HttpExecutorImpl(HttpClient client) {
...
+            try {
+                List<Organization> organizations = getOrganizations(response);
+                return new SearchResponse<Organization>(uri)
+                        .withHits(organizations)
+                        .withSize(getCount(response, organizations));
+            } catch (JsonProcessingException e) {
+                throw new FailedHttpRequestException(e.getMessage());
+            }
+        } else {
+            throw new NotFoundException(String.format(NOT_FOUND_MESSAGE_TEMPLATE, uri));
```
torarnet, line 88: Is this the only non successful case?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -52,11 +67,47 @@ public HttpExecutorImpl(HttpClient client) {
...
+                return new SearchResponse<Organization>(uri)
+                        .withHits(organizations)
+                        .withSize(getCount(response, organizations));
+            } catch (JsonProcessingException e) {
+                throw new FailedHttpRequestException(e.getMessage());
+            }
+        } else if (response.statusCode() == HTTP_NOT_FOUND) {
+            throw new NotFoundException(String.format(NOT_FOUND_MESSAGE_TEMPLATE, uri));
+        } else {
+            throw new FailedHttpRequestException(errorMessage(response));
```
torarnet, line 88: Is this error message the one that gets returned to client?

svenngjeto: It is beeing rewrappet in the handler, but the message is kept

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
svenngjeto: From cristin we have 404 (Not found, which may be considered a success) and 400 which is error in query statement - and  it might also be any other error from the net

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -52,11 +67,47 @@ public HttpExecutorImpl(HttpClient client) {
...
+    protected SearchResponse<Organization> query(URI uri) throws NotFoundException, FailedHttpRequestException {
+        HttpResponse<String> response = sendRequestMultipleTimes(addLanguage(uri)).get();
+        if (isSuccessful(response.statusCode())) {
+            try {
+                List<Organization> organizations = getOrganizations(response);
+                return new SearchResponse<Organization>(uri)
+                        .withHits(organizations)
+                        .withSize(getCount(response, organizations));
+            } catch (JsonProcessingException e) {
+                throw new FailedHttpRequestException(e.getMessage());
```
torarnet, line 83: Is json exception returned to client? Should we return something else?

svenngjeto: We go for json-ld as the default

torarnet: What I mean is: I see the statuscode is Bad Gateway, but what about the message back to client? Should we return the content of a JsonProcessingException back to the client, or should we hide the internal error and return a more generic one to the client?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
torarnet: So the client will receive a status code and message from cristin (from the Response object) and not from nva if this fails?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -52,11 +67,47 @@ public HttpExecutorImpl(HttpClient client) {
...
+            throw new FailedHttpRequestException(errorMessage(response));
+        }
+    }
+
+
+    private Organization getParentOrganization(SubSubUnitDto subSubUnitDto) {
+        URI parent = Optional.ofNullable(subSubUnitDto.getParentUnit()).map(InstitutionDto::getUri).orElse(null);
+        final Set<Organization> partOf = isNull(parent)
+                ? null
+                : Set.of(getParentOrganization(attempt(() ->
```
torarnet, line 97: Can this go into an infinite loop?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
```diff
@@ -52,11 +67,47 @@ public HttpExecutorImpl(HttpClient client) {
...
+                getSubSubUnitDtoWithMultipleEfforts(parent)).orElseThrow()));
+        return new Organization.Builder()
+                .withId(getNvaApiId(subSubUnitDto.getId(), ORGANIZATION_PATH))
+                .withPartOf(partOf)
+                .withName(subSubUnitDto.getUnitName()).build();
+    }
+
     private Organization fromSubSubunit(SubSubUnitDto subSubUnitDto) {
```
torarnet, line 105: Very similar to `getParentOrganization` . Might not be a problem

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
svenngjeto: I read something else in the first place...  We should hide/camouflage the json exception as something more infomative

torarnet (commented):
See other comments, will wait for response before approval.

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
svenngjeto: Only if there is a loop in the data in cristin

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
svenngjeto: It is a tiny difference in getting the siblings children - they are not so interesting when going upwards in the tree

cristin-organization/src/main/java/no/unit/nva/cristin/organization/HttpExecutorImpl.java
torarnet: Good thing we have a timeout in lambda

svenngjeto: I runs out of memory, have tried that....

torarnet (approved):

brinxmat (approved):

PR #116, opened 2022-01-12
Np 3530 open api post person

torarnet:
Part of a bigger feature for creating a person in cristin. Contains the OpenApi spec and a dummy handler implementation that was needed for SAM to build the stack using OpenApi spec


torarnet:
Most of deleted lines is because of deletion of person swagger spec which was not in use. The file person-swagger that was deleted is already implemented in cristin-proxy-swagger

svenngjeto (dismissed):

brinxmat (approved):

axthosarouris (approved):

PR #118, opened 2022-01-17
NP-3539: Model class for POST person

torarnet:


cristin-commons/src/main/java/no/unit/nva/cristin/common/Utils.java
```diff
@@ -40,4 +43,12 @@ public static boolean isPositiveInteger(String str) {
     public static boolean isOrcid(String identifier) {
         return Constants.ORCID_PATTERN.matcher(identifier).matches();
     }
+
+    /**
+     * A function that can be used to filter out duplicate values based on a given key.
+     */
+    public static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor) {
+        Map<Object, Boolean> seen = new ConcurrentHashMap<>();
```
brinxmat, line 51: A question about posting pre-existing NINs to Cristin: does Cristin pre-filter existing records with the same NIN, or can multiple records have the same NIN (not sure how the model is…)?

brinxmat (approved):

cristin-commons/src/main/java/no/unit/nva/cristin/common/Utils.java
torarnet: NIN must be unique to one person if I remember correctly

svenngjeto (approved):

PR #121, opened 2022-01-19
Np 3542 model person institution

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/PersonInstitutionInfo.java
```diff
@@ -0,0 +1,81 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import java.net.URI;
+import java.util.Objects;
+import nva.commons.core.JacocoGenerated;
+import nva.commons.core.JsonSerializable;
+
+@JacocoGenerated
+@JsonPropertyOrder({CONTEXT, ID, EMAIL, PHONE})
```
axthosarouris, line 16: Why is the order of the fields important?

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPersonInstitutionInfo.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import java.net.URI;
+import no.unit.nva.cristin.person.model.nva.PersonInstitutionInfo;
+import nva.commons.core.JacocoGenerated;
+
+@SuppressWarnings("unused")
+@JacocoGenerated
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
```
axthosarouris, line 11: Is this because this is the way the fields are sent by Cristin?

axthosarouris (commented):
Looks OK except for the field order

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/PersonInstitutionInfo.java
torarnet: Just thought we wanted context on top. If you think it is unneccesary i can remove the ordering

svenngjeto (dismissed):

brinxmat (approved):

PR #125, opened 2022-01-24
Np 3560 handler client post person

torarnet:
Will add authenticated cristin client later in review, and also remove commented karate action before merge. If this PR is too large, I can try to split it into smaller PR, but then I have to use more time to do so


cristin-person/src/main/java/no/unit/nva/cristin/person/create/CreateCristinPersonHandler.java
```diff
@@ -0,0 +1,113 @@
...
+        return DEFAULT_RESPONSE_MEDIA_TYPES;
+    }
+
+    @Override
+    protected Integer getSuccessStatusCode(Person input, Person output) {
+        return HttpURLConnection.HTTP_CREATED;
+    }
+
+    private void validateContainsPayload(Person input) throws BadRequestException {
+        if (Objects.isNull(input)) {
```
brinxmat, line 76: Static import

brinxmat (changes requested):

svenngjeto (approved):

brinxmat (dismissed):

axthosarouris (approved):

PR #149, opened 2022-03-07
Np 3489 validate query

svenngjeto:
Added simple SQL injection detector


torarnet (commented):
I need someone else' opinion

cristin-commons/src/main/java/no/unit/nva/utils/SqlRecognizer.java
```diff
@@ -0,0 +1,43 @@
...
+            "-",
+            ",");
+
+    /**
+     * Checks if there are possible SQL statements in a string.
+     * @param str String to checked for SQL statements
+     * @return true if there are possible SQL statements in str
+     */
+    public static boolean containsSqlInjectionFragmentCandidate(String str) {
+        Set<String> queryFragments  = new HashSet(List.of(str.toLowerCase(Locale.getDefault()).split(" ")));
```
axthosarouris, line 39: split on whitespace maybe?

axthosarouris:
I am not sure how robust this detection is and whether this is our problem. But since it is set as an own class we can refactor it later

axthosarouris (approved):

PR #151, opened 2022-03-09
internal api for cristin service

axthosarouris:


template.yaml
```diff
@@ -333,3 +355,20 @@ Resources:
       RestApiId: !Ref NvaCristinProxyApi
       Stage: !Ref NvaCristinProxyApi.Stage
 
+  NvaCristinProxyInternalApiBasePathMapping:
```
torarnet, line 358: Will this crash with NvaCristinProxyBasePathMapping?

docs/cristin-proxy-internal-apidoc.yaml
```diff
@@ -0,0 +1,1074 @@
...
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NvaCristinFetchPersonUsingNinFunction.Arn}/invocations
+        httpMethod: POST
+        type: "AWS_PROXY"
+      tags:
+        - Person
+      summary: Query person by national identity number
+      description: Returns a single person matching nin query parameter
+      operationId: GetPersonByNIN
+      security:
+        - CognitoUserPool: ['https://api.nva.unit.no/scopes/backend']
```
torarnet, line 41: So this is where the magic happens? Have you tested it locally? Should you write karate tests for it? Do you know if it works?

torarnet (commented):
See comments

template.yaml
axthosarouris: fixed

svenngjeto (approved):

torarnet (approved):

PR #152, opened 2022-03-09
NP-6193: Serializer, model class and unit tests for it

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/PersonPatch.java
```diff
@@ -0,0 +1,71 @@
...
+@JsonInclude(ALWAYS)
+@JacocoGenerated
+public class PersonPatch {
+
+    private transient Optional<String> orcid;
+    private transient Optional<String> firstName;
+    private transient Optional<String> lastName;
+    private transient Optional<String> preferredFirstName;
+    private transient Optional<String> preferredLastName;
+    private transient Optional<Boolean> reserved;
```
brinxmat, line 22: I understand why this is done — other reviewers may not, probably worth a comment in the code

brinxmat (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/PersonPatch.java
torarnet: Will do in next PR

svenngjeto (approved):

PR #162, opened 2022-03-21
Np 6913 api client patch person

torarnet:


brinxmat (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/update/UpdateCristinPersonApiClient.java
```diff
@@ -0,0 +1,45 @@
...
+        super(client);
+    }
+
+    /**
+     * Updates a person in Cristin.
+     *
+     * @return An empty json if update was successful
+     * @throws ApiGatewayException if something went wrong that can be mapped to a client response
+     */
+    public String updatePersonInCristin(String personId, ObjectNode request) throws ApiGatewayException {
```
svenngjeto, line 29: This could be Void?

brinxmat (approved):

brinxmat (approved):

svenngjeto (approved):

PR #170, opened 2022-03-28
Np 7865 handler auth patch person affiliation

torarnet:


svenngjeto (dismissed):

cristin-person/src/test/java/no/unit/nva/cristin/person/employment/update/UpdatePersonEmploymentHandlerTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @BeforeEach
+    void setUp() {
+        context = mock(Context.class);
+        output = new ByteArrayOutputStream();
+        handler = new UpdatePersonEmploymentHandler(environment);
+    }
+
+    @Test
+    void shouldReturnNoContentResponseWhenClientIsAuthenticated() throws IOException {
+        GatewayResponse<Void> gatewayResponse = sendQuery(validPath, EMPTY_JSON);
```
brinxmat, line 45: Should be null with no content

brinxmat (approved):

cristin-person/src/test/java/no/unit/nva/cristin/person/employment/update/UpdatePersonEmploymentHandlerTest.java
torarnet: So null instead of Void?

torarnet: Void is an empty response

brinxmat (approved):

PR #171, opened 2022-03-29
Np 7788 milliseconds

svenngjeto:


cristin-person/src/test/java/no/unit/nva/cristin/person/employment/create/personEmploymentCreate.feature
```diff
@@ -15,6 +15,22 @@ Feature: API tests for Creation of Cristin Person Employments
     * def simpleUserToken = tokenGenerator.loginUser(simple_user_name, simple_user_password, cognitoUserpoolId, cognitoClientAppId)
     * def invalidToken = 'just-a-invalid-token-for-now'
     * def personIdentifier = '515114'
+     # 'Test Testesen' has id 538786:
```
torarnet, line 18: I think his name has changed

cristin-person/src/test/java/no/unit/nva/cristin/person/employment/create/personEmploymentCreate.feature
```diff
@@ -36,3 +52,10 @@ Feature: API tests for Creation of Cristin Person Employments
     When method POST
     Then status 403
     And match response.title == 'Forbidden'
+
+  Scenario: Create returns status 201 Created for timestamp with zero millis
```
torarnet, line 56: Does this work when running multiple times? I think Cristin does not allow multiple employments with same data. Maybe comment this and uncomment when deletion is in place or add delete after this?

torarnet (commented):
See comments

brinxmat (approved):

PR #174, opened 2022-03-31
Np 8361 search project with organization

svenngjeto:


docs/cristin-proxy-swagger.yaml
```diff
@@ -46,6 +46,14 @@ paths:
...
             pattern: '.+'
           example: 'Helse'
+        - name: organization
+          in: query
+          description: Search projects by organization id.
+          required: false
+          schema:
+            type: string
+            pattern: '.+'
+          example: 'https://api.dev.nva.aws.unit.no/cristin/organization/20202.0.0.0'
```
torarnet, line 56: Are you using an uri as a query parameter? Is it not better to only use the last part 20202.0.0.0? I dont think it is wise to use 2 uris in the same request

cristin-commons/src/main/java/no/unit/nva/utils/UriUtils.java
```diff
@@ -128,4 +132,32 @@ public static URI createNvaPositionId(String code) {
         return attempt(() -> new URI(positionBase.getScheme(), positionBase.getHost(), positionBase.getPath(), code))
                 .orElseThrow();
     }
+
+    /**
+     * Convenience method to encode a uri.
```
torarnet, line 137: to encode or decode?

torarnet (commented):
See comments.

docs/cristin-proxy-swagger.yaml
svenngjeto: It's the URI which is the full and unambiguous id which precisely and persistent identifies the Organization - and we don't have to operate on partial identifers. 
Maybe how and what to use for id should be discussed for the whole project?

cristin-commons/src/main/java/no/unit/nva/utils/UriUtils.java
svenngjeto: To decode or not to decode - that's the question...

torarnet (approved):

brinxmat (approved):

PR #179, opened 2022-04-06
Np 8978 list projects organization

svenngjeto:
List projects for organization by adding /projects to organization id - same backend as project search with organization id (uri)


template.yaml
```diff
@@ -232,6 +232,22 @@ Resources:
...
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: cristin-project
+      Handler: no.unit.nva.cristin.projects.ListCristinOrganizationProjectsHandler::handleRequest
+      Events:
+        NvaListOrganizationProjectsEvent:
+          Type: Api
+          Properties:
+            RestApiId: !Ref NvaCristinProxyApi
+            Path: /organization/{id}/projects
```
torarnet, line 245: in openapi you have called it `identifier` . Here it is `id` . Not sure if it matters

cristin-project/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -118,6 +126,30 @@ public NvaProject queryOneCristinProjectUsingIdIntoNvaProject(String id, String
...
+                .withLanguage(requestQueryParameters.get(LANGUAGE))
+                .withItemsPerPage(requestQueryParameters.get(NUMBER_OF_RESULTS))
+                .toURI();
+        HttpResponse<String> response = listProjects(cristinUri);
+        List<CristinProject> cristinProjects =
+                getEnrichedProjectsUsingQueryResponse(response, requestQueryParameters.get(LANGUAGE));
+        List<NvaProject> nvaProjects = mapValidCristinProjectsToNvaProjects(cristinProjects);
+        long endRequestTime = System.currentTimeMillis();
+
+        URI id = getServiceUri(new HashMap(requestQueryParameters)); // createIdUriFromParams(rewrapOrganizationUri(requestQueryParameters), PROJECT);
```
torarnet, line 145: remove comment in next iteration

template.yaml
svenngjeto: It should be identifier.. Id is the URI..

cristin-project/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -118,7 +126,30 @@ public NvaProject queryOneCristinProjectUsingIdIntoNvaProject(String id, String
                 .withHits(nvaProjects);
     }
 
+    public SearchResponse<NvaProject> listOrganizationProjects(Map<String, String> requestQueryParameters)
+            throws ApiGatewayException {
+
+        long startRequestTime = System.currentTimeMillis();
+        URI cristinUri = new CristinQuery()
```
torarnet, line 133: Maybe extract method, but not important

torarnet (dismissed):

PR #181, opened 2022-04-07
Np 9065 project patch definitions

svenngjeto:
Open API specification with samples and AWS Template for PATCH project  service

### DIFF
--- a/docs/cristin-proxy-swagger.yaml
+++ b/docs/cristin-proxy-swagger.yaml
@@ -1383,12 +1383,6 @@ components:
     ProjectPatch:
       type: object
       properties:
-        '@context':
-          $ref: '#/components/schemas/Context'
-        type:
-          type: string
-          pattern: 'Project'
-          description: The type of the returned object, always Project
         title:
           type: string
           pattern: '.+'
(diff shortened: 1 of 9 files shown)


docs/cristin-proxy-swagger.yaml
```diff
@@ -1937,6 +2052,22 @@ components:
           en: "Summary in an academic manner"
         popularScientificSummary:
           en: "popularScientificSummary popularScientificSummary popularScientificSummary"
+    ProjectPatchExample:
+      value:
+        title: 'Updated Title'
+        alternativeTitles:
+          - 'en': 'Updated Title in english'
+        startDate: '2022-10-09T00:00:00.001Z'
+        endDate: '2030-12-31T00:00:00.001Z'
```
torarnet, line 2061: Does it now work with 000Z instead of 001Z?

docs/cristin-proxy-swagger.yaml
```diff
@@ -1328,6 +1380,69 @@ components:
...
+          type: string
+          pattern: '^http:\/\/lexvo.org\/id\/iso639-3\/[a-z]{3}$'
+          description: The lexvo URI for the language of the Project
+        alternativeTitles:
+          type: array
+          description: Titles in alternative languages
+          items:
+            $ref: '#/components/schemas/Languages'
+        startDate:
+          type: string
```
torarnet, line 1400: can this be a date-time format?

docs/cristin-proxy-swagger.yaml
svenngjeto: Yes,  CustomInstantSerializer is supposed to take care of that -( when the model comes in place....)

docs/cristin-proxy-swagger.yaml
```diff
@@ -110,6 +110,58 @@ paths:
...
+              objectExample:
+                $ref: '#/components/examples/ProjectPatchExample'
+      responses:
+        204:
+          description: Successful update of resource
+          content:
+            application/json:
+              schema:
+                type: string
+                example: '{}'
```
torarnet, line 149: I think this is now not a empty json but just Void/null

torarnet (dismissed):

torarnet (approved):

PR #182, opened 2022-04-07
Np 9063 validate input patch employment

torarnet:


svenngjeto (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/employment/update/UpdatePersonEmploymentValidator.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private static void validatePositionCode(ObjectNode input) throws BadRequestException {
+        if (input.has(TYPE)) {
+            URI type = parseUriField(input, TYPE);
+            Optional<String> code = Employment.extractPositionCodeFromTypeUri(type);
+            if (code.isEmpty() || code.get().isBlank()) {
+                throw new BadRequestException(invalidFieldParameterMessage(TYPE));
+            }
+        }
```
brinxmat, line 48: The intention here is to accept a type if and only if it has content? So there are two valid states: no type and a populated type?

cristin-person/src/main/java/no/unit/nva/cristin/person/employment/update/UpdatePersonEmploymentValidator.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private static void validateAffiliation(ObjectNode input) throws BadRequestException {
+        if (input.has(ORGANIZATION)) {
+            URI organization = parseUriField(input, ORGANIZATION);
+            String orgId = UriUtils.extractLastPathElement(organization);
+            if (isNull(CristinOrganization.fromIdentifier(orgId))) {
+                throw new BadRequestException(invalidFieldParameterMessage(ORGANIZATION));
+            }
+        }
```
brinxmat, line 58: Ditto

brinxmat (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/employment/update/UpdatePersonEmploymentValidator.java
torarnet: Cristin does not accept null for this field so if we send a value in this field it has to be populated as well. We cannot accept null or empty

PR #183, opened 2022-04-19
Np 9067 patch input model validate

svenngjeto:
Validation of patch request for project. Verifying required fields are not deleted.


cristin-project/src/main/java/no/unit/nva/cristin/projects/CristinApiClient.java
```diff
@@ -303,4 +304,8 @@ private URI getServiceUri(Map<String, String> queryParameters) {
                 .addQueryParameters(queryParameters)
                 .getUri();
     }
+
+    public Void updateProjectInCristin(String projectId, ObjectNode cristinJson) {
```
brinxmat, line 308: Probably comment why this is no-op

cristin-project/src/main/java/no/unit/nva/cristin/projects/FetchOneCristinProject.java
```diff
@@ -71,11 +71,11 @@ protected Integer getSuccessStatusCode(Void input, NvaProject output) {
         return HttpURLConnection.HTTP_OK;
     }
 
-    private String getValidId(RequestInfo requestInfo) throws BadRequestException {
-        attempt(() -> Integer.parseInt(requestInfo.getPathParameter(ID)))
-                .orElseThrow(failure -> new BadRequestException(ERROR_MESSAGE_INVALID_PATH_PARAMETER_FOR_ID));
+    public static String getValidIdentifier(RequestInfo requestInfo) throws BadRequestException {
+        attempt(() -> Integer.parseInt(requestInfo.getPathParameter(IDENTIFIER)))
```
brinxmat, line 75: I'm assuming that attempt catches both NPE and Number error, I would be tempted to return the value of the declarative code, but this involves a weird stringification at the end.

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+import static no.unit.nva.cristin.model.JsonPropertyNames.POPULAR_SCIENTIFIC_SUMMARY;
+import static no.unit.nva.cristin.model.JsonPropertyNames.START_DATE;
+import static no.unit.nva.cristin.model.JsonPropertyNames.TITLE;
+import static no.unit.nva.language.LanguageConstants.UNDEFINED_LANGUAGE;
+import static no.unit.nva.language.LanguageMapper.getLanguageByIso6391Code;
+import static no.unit.nva.language.LanguageMapper.getLanguageByUri;
+import static nva.commons.core.attempt.Try.attempt;
+
+
+public class ProjectPatchValidator {
```
brinxmat, line 32: A lot of the code in this class is repeated; I think there is a missed abstraction

template.yaml
```diff
@@ -155,7 +155,7 @@ Resources:
           Type: Api
           Properties:
             RestApiId: !Ref NvaCristinProxyApi
-            Path: /project/{id}
+            Path: /project/{identifier}
```
torarnet, line 158: Will fetching one project still work if you change variable name?

brinxmat (changes requested):

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private static void validateLanguageUri(JsonNode jsonNode) throws BadRequestException {
+        if (getLanguageByUri(URI.create(jsonNode.textValue())).equals(UNDEFINED_LANGUAGE)) {
+            throw new BadRequestException(String.format(ILLEGAL_VALUE_FOR_PROPERTY, LANGUAGE));
+        }
+    }
+
+    private static void validateInstant(ObjectNode input, String propertyName) throws BadRequestException {
+        attempt(() -> Instant.parse(input.get(propertyName).asText()))
```
torarnet, line 132: What if `propertyName` is not present? Will a missing enddate throw an exception?

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+    private static void validateRequiredFieldsNotNull(ObjectNode input) throws BadRequestException {
+        validateNotNullIfPresent(input, TITLE);
+        validateNotNullIfPresent(input, START_DATE);
+        validateNotNullIfPresent(input, CONTRIBUTORS);
+        validateNotNullIfPresent(input, COORDINATING_INSTITUTION);
+    }
+
+    private static void validateContributors(ObjectNode input) throws BadRequestException {
+        TypeReference<List<NvaContributor>> typeRef = new TypeReference<>() {
+        };
```
torarnet, line 65: Why the extra line?

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            final String content = input.get(CONTRIBUTORS).asText();
+            List<NvaContributor> contributors = OBJECT_MAPPER.readValue(content, typeRef);
+            if (contributors.isEmpty()) {
+                throw new BadRequestException(String.format(FIELD_CAN_NOT_BE_ERASED, CONTRIBUTORS));
+            }
+        } catch (JsonProcessingException e) {
```
torarnet, line 72: can this be replaced by an attempt? I think this is only relevant for `OBJECT_MAPPER.readValue`

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+            throw new BadRequestException(String.format(ILLEGAL_VALUE_FOR_PROPERTY, CONTRIBUTORS));
+        }
+    }
+
+    private static void validateStatus(ObjectNode input) throws BadRequestException {
+        validateNotNullIfPresent(input, no.unit.nva.cristin.model.JsonPropertyNames.STATUS);
```
torarnet, line 78: static import

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+        ProjectStatus.isValidStatus(input.get(no.unit.nva.cristin.model.JsonPropertyNames.STATUS).asText());
```
torarnet, line 79: This is a boolean. You should throw something if this is false

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+            List<NvaContributor> contributors = OBJECT_MAPPER.readValue(content, typeRef);
+            if (contributors.isEmpty()) {
+                throw new BadRequestException(String.format(FIELD_CAN_NOT_BE_ERASED, CONTRIBUTORS));
+            }
+        } catch (JsonProcessingException e) {
+            throw new BadRequestException(String.format(ILLEGAL_VALUE_FOR_PROPERTY, CONTRIBUTORS));
+        }
+    }
+
+    private static void validateStatus(ObjectNode input) throws BadRequestException {
```
torarnet, line 77: https://api.cristin-test.uio.no/v2/doc/json-schemas/projects_PATCH_request.json

Status is not a valid patch field. It gets calculated from dates

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+
+    public static final String FIELD_CAN_NOT_BE_ERASED = "Field %s can not be erased";
+    public static final String ILLEGAL_VALUE_FOR_PROPERTY = "Illegal value for '%s'";
+
+    /**
+     * Validate changes to Project, both nullable fields and values.
+     * @param input ObjectNode containing fields with input data to be changed.
+     * @throws BadRequestException thrown when input has illegal or invalid values.
+     */
+    public static void validate(ObjectNode input) throws BadRequestException {
```
torarnet, line 42: I dont think you need to validate all these fields as they are not supported in patch projects in cristin

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+        validateAlternativeTitles(input);
+        validateLanguageStringMap(input, POPULAR_SCIENTIFIC_SUMMARY);
+        validateInstant(input, END_DATE);
+        validateInstant(input, START_DATE);
+        validateLanguage(input);
+        validateStatus(input);
+        validateContributors(input);
+    }
+
+    private static void validateRequiredFieldsNotNull(ObjectNode input) throws BadRequestException {
```
torarnet, line 56: Name can be a little hard to understand. They are not required for every request, but if they are in payload they cannot be nulled

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,141 @@
...
+        };
+        try {
+            Map<String, String> languageAndStringMap =
+                    OBJECT_MAPPER.readValue(input.get(propertyName).asText(), typeRef);
+            for (String languageCode : languageAndStringMap.keySet()) {
+                validateLanguageCode(languageCode);
+                validateLanguageContent(languageAndStringMap.get(languageCode), propertyName);
+            }
+        } catch (JsonProcessingException e) {
+            throw new BadRequestException(String.format(ILLEGAL_VALUE_FOR_PROPERTY, propertyName));
```
torarnet, line 104: use attempt, extract method to be used by everything that throws this exception

torarnet (changes requested):
Some of these fields are not supported in Cristin Project PATCH.
Title is missing, coordinating institution is missing, participating institutions is missing.

template.yaml
svenngjeto: It's supposed to be renamed all the way- also when getting parameter from RequestInfo

cristin-project/src/main/java/no/unit/nva/cristin/projects/ProjectPatchValidator.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private static void validateJsonReadable(TypeReference<?> typeRef, String content, String propertyName)
+            throws BadRequestException {
+        attempt(() -> OBJECT_MAPPER.readValue(content, typeRef))
+                .orElseThrow(fail ->
+                        new BadRequestException(String.format(ILLEGAL_VALUE_FOR_PROPERTY, propertyName)));
+    }
+
+    private static void validateAlternativeTitles(ObjectNode input) throws BadRequestException {
+        validateNotNullIfPresent(input,ALTERNATIVE_TITLES);
```
torarnet, line 64: Can we erase alternative titles by sending null or empty list? Since that field is not supported at present in cristin we can think about that later. Seems good for now

torarnet (commented):
Have you tested an empty json with only one valid field and checked if the others are ignored? I think you should do that to verify that we dont have any required fields in patch (or just do an unit test with empty object before merge)

Also create nullable versions of validators so we can differ between non nullable properties and nullable properties and validate both.

brinxmat (changes requested):
See other reviewer's comments

torarnet (approved):
Approved but add validation of title later

brinxmat (approved):

PR #186, opened 2022-04-27
Np 9087 user can change orcid

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/update/CristinPersonPatchJsonCreator.java
```diff
@@ -42,6 +42,15 @@ public CristinPersonPatchJsonCreator create() {
         return this;
     }
 
+    /**
+     * Create Cristin json object from values allowed to be changed by the user himself.
```
brinxmat, line 46: himself -> themselves

cristin-person/src/main/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandler.java
```diff
@@ -44,16 +50,51 @@ protected Void processInput(String input, RequestInfo requestInfo, Context conte
...
+
+    private void checkHasFields(ObjectNode cristinJson) throws BadRequestException {
         if (cristinJson.isEmpty()) {
             throw new BadRequestException(ERROR_MESSAGE_NO_SUPPORTED_FIELDS_IN_PAYLOAD);
         }
+    }
 
-        String personId = getValidPersonId(requestInfo);
+    private void checkIdentifiersMatch(String personId, String personIdFromCognito) throws BadRequestException {
+        if (!Objects.equals(personId, personIdFromCognito)) {
```
brinxmat, line 82: Maybe extract this as a method since it's difficu!t to see hidden exc!amation marks

cristin-person/src/test/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandlerTest.java
```diff
@@ -187,8 +196,44 @@ void shouldReturnBadRequestWhenInvalidIdentifier() throws IOException {
         assertThat(gatewayResponse.getBody(), containsString(ErrorMessages.invalidPathParameterMessage(PERSON_ID)));
     }
 
-    private static String randomIntegerAsString() {
-        return String.valueOf(randomInteger());
+    @Test
+    void shouldReturnNoContentResponseWhenCallingHandlerWithAllowedJsonValuesAndUserIsAuthenticatedAsHimself()
```
brinxmat, line 200: AsHimself — is this necessary? Is the user not just authenticated? Either way, use gender-non-specific themselves. Perhaps this is more to do with matching user in token and elsewhere? Not sure what the intent is

cristin-person/src/test/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandlerTest.java
```diff
@@ -187,8 +196,44 @@ void shouldReturnBadRequestWhenInvalidIdentifier() throws IOException {
...
+        ObjectNode jsonObject = OBJECT_MAPPER.createObjectNode();
+        jsonObject.put(ORCID, VALID_ORCID);
+        GatewayResponse<Void> gatewayResponse = queryWithPersonCristinIdButNoAccessRights(validPath,
+                                                                                          jsonObject.toString());
+
+        assertEquals(HttpURLConnection.HTTP_NO_CONTENT, gatewayResponse.getStatusCode());
+    }
+
+    @Test
+    void shouldThrowBadRequestWhenUserIsAuthenticatedAsHimselfButDoesNotSendAnyFieldsHeIsAllowedToUpdate()
```
brinxmat, line 211: Ditto

brinxmat (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandler.java
torarnet: It is almost the only thing in this method. Should i extract it to another method? Do you have a suggestion on how to make the exclamation mark more visible?

cristin-person/src/test/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandlerTest.java
torarnet: A user can either act as themselves, or as an user administrator. Therefore i specify themselves to differentiate between changing own data and changing others data

cristin-person/src/main/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandler.java
torarnet:
````
private boolean notEqual(String personId, String personIdFromCognito) {
    return !Objects.equals(personId, personIdFromCognito);
}
````

brinxmat (approved):

svenngjeto (approved):

PR #187, opened 2022-04-29
Updated Positions API to filter out status based on active query para…

RanjithTom:
…m in the request


cristin-person/src/main/java/no/unit/nva/cristin/person/affiliations/handler/PositionCodesHandler.java
```diff
@@ -35,11 +35,16 @@ public PositionCodesHandler(CristinPositionCodesClient apiClient, Environment en
...
     }
 
     @Override
     protected Integer getSuccessStatusCode(Void input, PositionCodes output) {
         return HttpURLConnection.HTTP_OK;
     }
+
+    private Boolean extractPositionCodeStatusLookingFor(RequestInfo requestInfo) {
+        return requestInfo.getQueryParameterOpt(ACTIVE_STATUS_QUERY_PARAM)
+            .map(Boolean::valueOf).orElse(null);
```
torarnet, line 48: Does this work if the value is something else than boolean? What if i add param `active=hello` ?

PR #188, opened 2022-04-29
Np 9070 positions query

RanjithTom:


cristin-person/src/main/java/no/unit/nva/cristin/person/affiliations/handler/PositionCodesHandler.java
```diff
@@ -35,11 +41,28 @@ public PositionCodesHandler(CristinPositionCodesClient apiClient, Environment en
...
     }
 
     @Override
     protected Integer getSuccessStatusCode(Void input, PositionCodes output) {
         return HttpURLConnection.HTTP_OK;
     }
+
+    private Boolean extractPositionCodeStatusQueryParam(RequestInfo requestInfo) throws BadRequestException {
+        var activeStatusQueryParam = requestInfo.getQueryParameterOpt(ACTIVE_STATUS_QUERY_PARAM)
+            .map(String::valueOf).orElse(null);
```
torarnet, line 54: This is already a String. I think doing `String::valueOf` is unnecessary.

RanjithTom: This is Optional string so we need to map this to a String value

torarnet: What about:
````
return requestInfo.getQueryParameterOpt(ACTIVE_STATUS_QUERY_PARAM)
    .map(this::validateAndGetPositionCodeStatusBooleanQueryParam).orElse(null);
````

torarnet (commented):
Have commented a suggestion you should look at

cristin-person/src/main/java/no/unit/nva/cristin/person/affiliations/handler/PositionCodesHandler.java
torarnet: Optional gets unwrapped when you do the `orElse`

RanjithTom: > What about:
> 
> ```
> return requestInfo.getQueryParameterOpt(ACTIVE_STATUS_QUERY_PARAM)
>     .map(this::validateAndGetPositionCodeStatusBooleanQueryParam).orElse(null);
> ```
validateAndGetPositionCodeStatusBooleanQueryParam throws BadRequestException so it's not able to define in method reference inside map

Will Update code as per simply using `orElse `for unwrapping string

torarnet (dismissed):

axthosarouris (changes requested):
Good job. Replace with var where applicable

PR #199, opened 2022-06-03
Np 9150 person query per organization

torarnet:


.gitignore
```diff
@@ -288,3 +288,4 @@ gradle-app.setting
 
 /cristin-project/target/
 /cristin-organization/target/
+/cristin-person/target/
```
brinxmat, line 291: Is there something like /*/target that can be used to generalise this?

brinxmat (approved):

axthosarouris (approved):

PR #200, opened 2022-06-07
Np 9150 sort persons based on identifier

torarnet:


svenngjeto (approved):

torarnet:
Since we are sorting only a subset of results and not the full resultset from upstream we will do it another way. Will maybe close this PR

PR #211, opened 2022-08-31
Np 9239 put person with employments

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/update/PersonPatchValidator.java
```diff
@@ -50,4 +63,19 @@ private static void validateReservedIfPresent(ObjectNode input) throws BadReques
             throw new BadRequestException(RESERVED_FIELD_CAN_ONLY_BE_SET_TO_TRUE);
         }
     }
+
+    private static void validateEmploymentsIfPresent(ObjectNode input) throws BadRequestException {
```
brinxmat, line 67: I'd probably add the following just to make it more readable. Note, use the correct printf-message formatting in the EXCEPTION_WHEN_VALIDATING_EMPLOYMENTS, i.e. with {}.

```
    private static void validateEmploymentsIfPresent(ObjectNode input) throws BadRequestException {
        if (input.has(EMPLOYMENTS) && !input.get(EMPLOYMENTS).isNull()) {
                var employments = parseEmployments(input);
                for (Employment employment : employments) {
                    CreatePersonEmploymentValidator.validate(employment);
                }
        }
    }

    private static Employment[] parseEmployments(ObjectNode input) throws BadRequestException {
        try {
            return OBJECT_MAPPER.readValue(input.get(EMPLOYMENTS).toString(), Employment[].class);
        } catch (JsonProcessingException e) {
            logger.warn(EXCEPTION_WHEN_VALIDATING_EMPLOYMENTS, e.getMessage());
            throw new BadRequestException(COULD_NOT_PARSE_EMPLOYMENT_FIELD);
        }
    }
    
```

cristin-person/src/main/java/no/unit/nva/cristin/person/update/PersonPatchValidator.java
```diff
@@ -1,19 +1,31 @@
...
 
+    private static final Logger logger = LoggerFactory.getLogger(PersonPatchValidator.class);
+
     public static final String ORCID_IS_NOT_VALID = "ORCID is not valid";
     public static final String FIELD_CAN_NOT_BE_ERASED = "Field %s can not be erased";
     public static final String RESERVED_FIELD_CAN_ONLY_BE_SET_TO_TRUE =
         "Reserved field can only be set to true if present";
+    public static final String COULD_NOT_PARSE_EMPLOYMENT_FIELD = "Could not parse employment field because of "
+                                                                  + "invalid data";
+    private static final String EXCEPTION_WHEN_VALIDATING_EMPLOYMENTS = "Exception when validating employments: ";
```
brinxmat, line 28: Here, you want {}, see below

cristin-person/src/main/java/no/unit/nva/cristin/person/update/PersonPatchValidator.java
```diff
@@ -23,6 +35,7 @@ public static void validate(ObjectNode input) throws BadRequestException {
         validateFirstNameIfPresent(input);
         validateLastNameIfPresent(input);
         validateReservedIfPresent(input);
+        validateEmploymentsIfPresent(input);
```
brinxmat, line 38: Add a private constructor since this is all static and the implicit constructor from Java should not be accessible:

@JacocoGenerated
private PersonPatchValidator() {
    // NO-OP
}

brinxmat (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPerson.java
```diff
@@ -213,6 +214,9 @@ private Set<Affiliation> extractAffiliations() {
     }
 
     private Set<Employment> extractEmployments() {
+        if (isNull(getDetailedAffiliations())) {
```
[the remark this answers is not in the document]
torarnet: We do not want to default to empty list because if we PUT PATCH that back we delete all affiliations

brinxmat (approved):

PR #213, opened 2022-09-01
Np 10007 remove old nin field

torarnet:


torarnet:
Codacy is only failing because target branch is not main for the time being

brinxmat (approved):

brinxmat (approved):

axthosarouris (approved):

PR #218, opened 2022-09-26
Np 12276 inst header to cristin

torarnet:
Dependant on new version of Cristin Api which is not out yet but proxy should still work, for the time being the new version of Cristin Api should be backwards compatible to prevent breaking changes. A little work in progress but want feedback

### DIFF
--- a/cristin-commons/src/main/java/no/unit/nva/cristin/model/Constants.java
+++ b/cristin-commons/src/main/java/no/unit/nva/cristin/model/Constants.java
@@ -61,6 +61,7 @@ public class Constants {
     public static final String POSITION = "position";
     public static final String HASHTAG = "#";
     public static final String SLASH_DELIMITER = "/";
+    public static final String CRISTIN_INSTITUTION_HEADER = "Cristin-Institution";
 
     public enum QueryType {
         QUERY_USING_GRANT_ID,
--- a/cristin-person/src/main/java/no/unit/nva/cristin/person/create/CreateCristinPersonApiClient.java
+++ b/cristin-person/src/main/java/no/unit/nva/cristin/person/create/CreateCristinPersonApiClient.java
@@ -26,10 +26,10 @@ public class CreateCristinPersonApiClient extends PostApiClient {
     /**
      * Used for creating a person in Cristin from the supplied Person object.
      */
-    public Person createPersonInCristin(Person person) throws ApiGatewayException {
+    public Person createPersonInCristin(Person person, String instNr) throws ApiGatewayException {
         String payload = generatePayloadFromRequest(person);
         URI uri = getCristinPersonPostUri();
-        HttpResponse<String> response = post(uri, payload);
+        HttpResponse<String> response = post(uri, payload, instNr);
         checkPostHttpStatusCode(getNvaApiUri(PERSON_PATH_NVA), response.statusCode());
 
         return createPersonFromResponse(response);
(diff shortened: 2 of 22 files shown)


cristin-commons/src/main/java/no/unit/nva/cristin/model/Constants.java
```diff
@@ -61,6 +61,7 @@ public class Constants {
     public static final String POSITION = "position";
     public static final String HASHTAG = "#";
     public static final String SLASH_DELIMITER = "/";
+    public static final String CRISTIN_INSTITUTION_HEADER = "Cristin-Institution";
```
brinxmat, line 64: We should indicate what the value means in context, so possibly something like "On-Behalf-Of-Institution", if possible to change this (it gives more meaning in context and says what it does).

On a side note, X- prefixes, which I half expected to see here, were deprecated in 2012…the things you learn doing PRs.

cristin-person/src/main/java/no/unit/nva/cristin/person/create/CreateCristinPersonApiClient.java
```diff
@@ -35,6 +35,18 @@ public Person createPersonInCristin(Person person) throws ApiGatewayException {
         return createPersonFromResponse(response);
     }
 
+    /**
+     * Used for creating a person in Cristin from the supplied Person object and at allowed Cristin institution.
```
brinxmat, line 39: Used for creating -> Creates

cristin-person/src/main/java/no/unit/nva/cristin/person/employment/create/CreatePersonEmploymentClient.java
```diff
@@ -29,15 +29,16 @@ public CreatePersonEmploymentClient(HttpClient httpClient) {
      *
      * @param identifier Identifier of Person to add employment to
      * @param employment Request object to serialize and send to upstream
+     * @param instNr First number sequence of a Cristin organization identifier
      * @return the response from upstream deserialized to object
      * @throws ApiGatewayException if something went wrong and that can be returned to client
      */
-    public Employment createEmploymentInCristin(String identifier, Employment employment)
+    public Employment createEmploymentInCristin(String identifier, Employment employment, String instNr)
```
brinxmat, line 36: call the argument cristinInstitutionNumber…we need a paper trail of what this is (you document it as such in the comment, but it oddly isn't the first place we look and folk are lazy)

cristin-person/src/test/java/no/unit/nva/cristin/person/create/CreateCristinPersonHandlerTest.java
```diff
@@ -67,6 +73,9 @@ public class CreateCristinPersonHandlerTest {
     private static final String DUMMY_FIRST_NAME = randomString();
     private static final String DUMMY_LAST_NAME = randomString();
     private static final String DUMMY_CRISTIN_ID = "123456";
+    private static final String ANOTHER_ORGANIZATION = "https://api.dev.nva.aws.unit.no/cristin/organization/20202.0.0"
```
brinxmat, line 76: Here, you're using real examples, but you could test with https://example.org/20000.x.y.z, since we parse the number uniformly as the last path element

cristin-person/src/test/java/no/unit/nva/cristin/person/employment/create/CreatePersonEmploymentHandlerTest.java
```diff
@@ -152,4 +153,17 @@ private GatewayResponse<Employment> queryWithoutRequiredAccessRights() throws IO
 
         return GatewayResponse.fromOutputStream(output, Employment.class);
     }
+
+    private GatewayResponse<Employment> sendQueryWithCristinOrgId(Employment body) throws IOException {
+        var customerId = randomUri();
+        InputStream input = new HandlerRequestBuilder<Employment>(OBJECT_MAPPER)
```
brinxmat, line 159: var

brinxmat (commented):
In general, this seems fine. I think we will re-visit this at a later point, but it is important that it works now. If there is anything specific that you want to talk about, I am available :)

brinxmat (approved):

PR #219, opened 2022-09-27
NP-13344: Split person pojo creation between auth and non auth

torarnet:


axthosarouris (commented):
The code looks good to me, but I am not familiar with the context of the changes.

brinxmat (approved):

brinxmat (approved):

PR #229, opened 2022-11-01
Np 16240 total funding amount

torarnet:


cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/CristinFundingAmount.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @JsonCreator
+    public CristinFundingAmount(@JsonProperty(CRISTIN_CURRENCY_CODE) String currencyCode,
+                                @JsonProperty(AMOUNT) Double amount) {
+        this.currencyCode = currencyCode;
+        this.amount = amount;
+    }
+
+    public String getCurrencyCode() {
+        return currencyCode;
```
brinxmat, line 28: Is the currency code an ISO4217 code?

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/FundingAmount.java
```diff
@@ -0,0 +1,59 @@
...
+    @JsonProperty(AMOUNT)
+    private final transient Double amount;
+
+    @JsonCreator
+    public FundingAmount(@JsonProperty(CURRENCY_CODE) String currencyCode, @JsonProperty(AMOUNT) Double amount) {
+        this.currencyCode = currencyCode;
+        this.amount = amount;
+    }
+
+    public String getCurrencyCode() {
```
brinxmat, line 28: Probably just "currency"

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/FundingAmount.java
```diff
@@ -0,0 +1,59 @@
...
+        return currencyCode;
+    }
+
+    public Double getAmount() {
```
brinxmat, line 32: probably "value"

brinxmat (dismissed):

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/CristinFundingAmount.java
torarnet: The string "NOK" is hardcoded in the Cristin model. We have to change that when we add support for creation and update of funding amount

brinxmat (approved):

PR #231, opened 2022-11-04
Np 16240 project categories

torarnet:


cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/TypedLabel.java
```diff
@@ -0,0 +1,58 @@
...
+    public TypedLabel(@JsonProperty(TYPE) String type, @JsonProperty(LABELS) Map<String, String> labels) {
+        this.type = type;
+        this.labels = labels;
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public Map<String, String> getLabels() {
```
brinxmat, line 31: Call it "label", reserve the plural form for arrays

cristin-project/src/test/java/no/unit/nva/cristin/projects/NvaProjectBuilderTest.java
```diff
@@ -90,4 +97,14 @@ void shouldSerializeAndDeserializeFundingAmountIntoSameObject() {
         assertThat(deserialized, equalTo(fundingAmount));
     }
 
+    @Test
+    void shouldSerializeAndDeserializeTypedLabelIntoSameObject() {
```
brinxmat, line 101: Remove these tests and keep a single top-level (i.e. the root class for each data structure) that ensures that the entire tree can be built (it will cover all these cases too)

brinxmat (changes requested):

cristin-project/src/test/java/no/unit/nva/cristin/projects/NvaProjectBuilderTest.java
torarnet: Will do in separate PR. It requires some work on the model classes

brinxmat (approved):

PR #235, opened 2022-11-09
Np 16240 field related projects

torarnet:


cristin-project/src/test/java/no/unit/nva/cristin/projects/project.feature
```diff
@@ -99,80 +99,42 @@ Feature: API tests for Cristin Project retrieve and search
     * def uri = 'https://api.dev.nva.aws.unit.no/' + basePath + '/project/0'
     And match response.detail == "The requested resource '" + uri + "' was not found"
 
-  Scenario: Fetch returns all supported fields
-    Given path '/project/6720765'
+  Scenario: Fetch returns supported fields
```
torarnet, line 102: Reducing karate code as we can test a lot of fields using only one project from identifier

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/NvaProject.java
```diff
@@ -35,12 +21,8 @@
 
 @SuppressWarnings({"PMD.ExcessivePublicCount","PMD.TooManyFields"})
 @JsonInclude(ALWAYS)
-@JsonPropertyOrder({CONTEXT, ID, TYPE, IDENTIFIERS, TITLE, LANGUAGE, ALTERNATIVE_TITLES, START_DATE, END_DATE,
```
torarnet, line 38: I do not see any reason to keep this ordering. I know it was needed for unit tests at some early stage, but now we have improved the tests and order of fields should not be important

brinxmat (approved):

PR #236, opened 2022-11-10
Np 16239 refactor project module structure

torarnet:


torarnet:
The main thing in this PR is that i moved the classes for Update and Create project into their own packages create and update

brinxmat (approved):

PR #239, opened 2022-11-30
NP-21545 bug related. Fixing map

tkardu:
Fixing Map.of(...) bug.

### DIFF
--- a/cristin-project/src/main/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandler.java
+++ b/cristin-project/src/main/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandler.java
@@ -15,6 +15,7 @@ import java.net.HttpURLConnection;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 
 import static no.unit.nva.cristin.common.ErrorMessages.ERROR_MESSAGE_INVALID_PATH_PARAMETER_FOR_ID_FOUR_NUMBERS;
@@ -100,12 +101,12 @@ public class QueryCristinOrganizationProjectHandler extends CristinQueryHandler<
         return cristinApiClient.listOrganizationProjects(requestQueryParameters);
     }
 
-    private Map<String, String> extractQueryParameters(RequestInfo requestInfo) throws BadRequestException {
-        var requestQueryParameters = Map.of(
+    private ConcurrentHashMap<String, String> extractQueryParameters(RequestInfo requestInfo) throws BadRequestException {
+        var requestQueryParameters = new ConcurrentHashMap<>(Map.of(
                 CRISTIN_QUERY_PARAMETER_PARENT_UNIT_ID, getValidId(requestInfo),
                 LANGUAGE, getValidLanguage(requestInfo),
                 PAGE, getValidPage(requestInfo),
-                NUMBER_OF_RESULTS, getValidNumberOfResults(requestInfo));
+                NUMBER_OF_RESULTS, getValidNumberOfResults(requestInfo)));
         //From here
         VALID_QUERY_PARAM_NO_VALIDATION.forEach(paramName -> putOrNotQueryParameterOrEmpty(requestInfo,
                 paramName, requestQueryParameters));
@@ -149,7 +150,7 @@ public class QueryCristinOrganizationProjectHandler extends CristinQueryHandler<
      * Putting the parameter if it exists in a map.
      */
     protected void putOrNotQueryParameterOrEmpty(RequestInfo requestInfo, String parameter,
-                                                 Map<String,String> requestQueryParameters) {
+                                                 ConcurrentHashMap<String,String> requestQueryParameters) {
         if (getQueryParameter(requestInfo, parameter).isPresent()) {
             requestQueryParameters.put(parameter,getQueryParameter(requestInfo, parameter).get());
         }


cristin-project/src/test/java/no/unit/nva/cristin/projects/query/projectQuery.feature
```diff
@@ -192,4 +192,15 @@ Feature: API tests for Cristin projects query
       | 'NOTSTARTED' |
       | 'ACTIVE' |
 
+  Scenario: Query with extended list of parameters and valid values returns OK
```
torarnet, line 195: Do the same in `projectOrganization.feature`

cristin-project/src/test/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandlerTest.java
```diff
@@ -131,6 +136,44 @@ void shouldReturnSearchResponseWithEmptyHitsWhenBackendFetchIsEmpty() throws Api
...
 
+    @Test
+    void shouldAddParamsToCristinQueryForFilteringAndReturnOk() throws IOException, ApiGatewayException {
+        cristinApiClient = spy(cristinApiClient);
+        doReturn(new HttpResponseFaker(EMPTY_LIST_STRING,
+                HttpURLConnection.HTTP_OK, generateHeaders(ZERO_VALUE, LINK_EXAMPLE_VALUE)))
+                .when(cristinApiClient).listProjects(any());
+        handler = new QueryCristinOrganizationProjectHandler(cristinApiClient, new Environment());
+        var queryParams = Map.of("funding", "NRE:1234",
+                "biobank", "123321",
```
torarnet, line 147: Import constant with value biobank and use `randomInteger()`

cristin-project/src/test/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandlerTest.java
```diff
@@ -131,6 +136,44 @@ void shouldReturnSearchResponseWithEmptyHitsWhenBackendFetchIsEmpty() throws Api
...
+                "keyword", "nature",
```
torarnet, line 148: `randomString()` and constant?

cristin-project/src/test/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandlerTest.java
```diff
@@ -131,6 +136,44 @@ void shouldReturnSearchResponseWithEmptyHitsWhenBackendFetchIsEmpty() throws Api
...
     }
 
+    @Test
+    void shouldAddParamsToCristinQueryForFilteringAndReturnOk() throws IOException, ApiGatewayException {
+        cristinApiClient = spy(cristinApiClient);
+        doReturn(new HttpResponseFaker(EMPTY_LIST_STRING,
+                HttpURLConnection.HTTP_OK, generateHeaders(ZERO_VALUE, LINK_EXAMPLE_VALUE)))
+                .when(cristinApiClient).listProjects(any());
+        handler = new QueryCristinOrganizationProjectHandler(cristinApiClient, new Environment());
+        var queryParams = Map.of("funding", "NRE:1234",
```
torarnet, line 146: I was advised to use strings as constant variables when started working in the project, but maybe you can reuse the names you are using in the query parameter map where you defined a lot of constants, instead of writing them again. Also, we have a library generating random data is I have described in other comment. I don't think we need exact matches for all query params since we are mocking upstream and just validating that the service accepts the params and puts them on URI

cristin-commons/src/main/java/no/unit/nva/cristin/model/JsonPropertyNames.java
```diff
@@ -42,6 +42,7 @@ public class JsonPropertyNames {
     public static final String ORGANIZATION = "organization";
     public static final String CRISTIN_UNIT_ID = "cristin_unit_id";
     public static final String CRISTIN_INSTITUTION_ID = "cristin_institution_id";
+    public static final String CRISTIN_QUERY_PARAMETER_LANGUAGE_KEY = "lang";
```
torarnet, line 45: We not already have this somewhere?

cristin-project/src/main/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandler.java
```diff
@@ -100,12 +102,13 @@ protected SearchResponse<NvaProject> processInput(Void input, RequestInfo reques
         return cristinApiClient.listOrganizationProjects(requestQueryParameters);
     }
 
-    private Map<String, String> extractQueryParameters(RequestInfo requestInfo) throws BadRequestException {
-        var requestQueryParameters = Map.of(
+    private ConcurrentHashMap<String, String> extractQueryParameters(RequestInfo requestInfo)
```
torarnet, line 105: I think you can keep the abstract type Map as it is, and rather only use the implementation when creating instance of it.

Example: `Map<String, String> map = new ConcurrentHashMap<>()`
Another: 
```
private Map<String, String> extractQueryParameters() {
  return new ConcurrentHashMap<>(with the data)
}  
```

cristin-project/src/main/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandler.java
```diff
@@ -149,7 +152,7 @@ protected void validateQueryParameterKeys(RequestInfo requestInfo) throws BadReq
      * Putting the parameter if it exists in a map.
      */
     protected void putOrNotQueryParameterOrEmpty(RequestInfo requestInfo, String parameter,
-                                                 Map<String,String> requestQueryParameters) {
+                                                 ConcurrentHashMap<String,String> requestQueryParameters) {
```
torarnet, line 155: I don't think changing type here is necessary, see other comment

torarnet (commented):
See comments. If you feel like some of them are not necessary leave a reply

cristin-project/src/test/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandlerTest.java
```diff
@@ -131,6 +143,42 @@ void shouldReturnSearchResponseWithEmptyHitsWhenBackendFetchIsEmpty() throws Api
...
+                "sort", START_DATE);
+        handler.handleRequest(generateHandlerProRealisticRequest(queryParams), output, context);
+        var captor = ArgumentCaptor.forClass(URI.class);
+
+        verify(cristinApiClient).listProjects(captor.capture());
+        var actualURI = captor.getValue().toString();
+        assertThat(actualURI,
+                containsString("page=5"));
+        assertThat(actualURI,
+                containsString("&biobank=" + BIOBANK_SAMPLE));
```
torarnet, line 167: For later: 
```
private static final String QUERY_PARAM_URI_TEMPLATE = "&%s=%s;

private static String formattedUriParam(String key, String value) {
  return String.format(QUERY_PARAM_URI_TEMPLATE, key, value)
}

...containsString(formattedUriParam(BIOBANK_ID, BIOBANK_SAMPLE)
```

Or maybe this one could be used, comparing the whole map in one assert?

https://github.com/BIBSYSDEV/nva-cristin-service/blob/main/cristin-commons/src/main/java/no/unit/nva/utils/UriUtils.java

Method `formatQueryParameters` , used by `queryParameters`

cristin-project/src/test/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandlerTest.java
```diff
@@ -131,6 +143,42 @@ void shouldReturnSearchResponseWithEmptyHitsWhenBackendFetchIsEmpty() throws Api
...
+    @Test
+    void shouldAddParamsToCristinQueryForFilteringAndReturnOk() throws IOException, ApiGatewayException {
+        cristinApiClient = spy(cristinApiClient);
+        doReturn(new HttpResponseFaker(EMPTY_LIST_STRING,
+                HttpURLConnection.HTTP_OK, generateHeaders(ZERO_VALUE, LINK_EXAMPLE_VALUE)))
+                .when(cristinApiClient).listProjects(any());
+        handler = new QueryCristinOrganizationProjectHandler(cristinApiClient, new Environment());
+        var queryParams = Map.of("funding", FUNDING_SAMPLE,
+                "biobank", BIOBANK_SAMPLE,
+                "keyword", KEYWORD_SAMPLE,
```
torarnet, line 155: For later, consider also using constants for the keys, i think you already have defined them here: https://github.com/BIBSYSDEV/nva-cristin-service/blob/main/cristin-commons/src/main/java/no/unit/nva/cristin/model/JsonPropertyNames.java

torarnet (dismissed):
Approved with comments. Please read them before merge and decide if you want to change anything or just merge

cristin-project/src/test/java/no/unit/nva/cristin/projects/query/organization/QueryCristinOrganizationProjectHandlerTest.java
torarnet: queryParameters method do a sort so the ordering should always be the same

PR #240, opened 2022-12-01
Swager changes. Putting it aside from other fixes. Might need recheck…

tkardu:
… type of levels, biobank and modified since


docs/cristin-proxy-swagger.yaml
```diff
@@ -88,6 +88,99 @@ paths:
             type: number
             default: '5'
           example: '10'
+        - name: institution
```
torarnet, line 91: In code, you are using `public static final String CRISTIN_INSTITUTION_ID = "cristin_institution_id";` as input param, but when adding to query you are using `private static final String CRISTIN_QUERY_PARAMETER_INSTITUTION = "institution";`

So either you should change input param to `institution` or update swagger to use `cristin_institution_id`

docs/cristin-proxy-swagger.yaml
```diff
@@ -520,6 +613,99 @@ paths:
...
```
torarnet, line 616: Same here as other comment

docs/cristin-proxy-swagger.yaml
```diff
@@ -88,6 +88,99 @@ paths:
...
+          required: false
+          schema:
+            type: string
+        - name: sort
+          in: query
+          description: here
+          required: false
+          schema:
+            type: string
+        - name: name
```
torarnet, line 133: What is `name` ?

docs/cristin-proxy-swagger.yaml
```diff
@@ -88,6 +88,99 @@ paths:
...
+          in: query
+          description: Sorts on 'start_date' and/or 'end_date'. See Search and sort. Default sort order is on 'cristin_project_id' in ascending order.
```
torarnet, line 135: I think this belongs under param `sort`

docs/cristin-proxy-swagger.yaml
```diff
@@ -520,6 +613,99 @@ paths:
...
```
torarnet, line 660: See other comment

docs/cristin-proxy-swagger.yaml
torarnet: `institution` is the param name in Cristin

docs/cristin-proxy-swagger.yaml
tkardu: Name was a dummy-template which I was feeling out for each parameter.
Fixed it now

tkardu: moved description for sort in sort

docs/cristin-proxy-swagger.yaml
tkardu: I will use institution for my purposes.

torarnet (approved):

PR #241, opened 2022-12-15
code ready

ketilaa:


cristin-funding/build.gradle
```diff
@@ -0,0 +1,43 @@
+dependencies {
+    implementation project(":cristin-commons")
+    implementation libs.nva.json
+    implementation libs.nva.core
+    implementation libs.nva.language
+//    implementation libs.aws.sdk2.core
```
torarnet, line 6: Remove commented deps?

cristin-funding/src/test/resources/success.json
```diff
@@ -0,0 +1,3972 @@
+[
+  {
+    "code": "EC/FP7",
```
torarnet, line 3: So this is where all the lines comes from. I thought I had a very big PR but 4/5 of lines is from this file. 

Do you need every entry in cristin? Does it load fast when testing?

template.yaml
```diff
@@ -473,6 +473,36 @@ Resources:
             Path: /person/{id}/employment/{employmentId}/
             Method: patch
 
+  NvaCristinListFundingSourcesFunction:
+    Description: Function for listing all funding sources
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: cristin-funding
+      Handler: no.unit.nva.cristin.funding.sources.query.ListFundingSourcesHandler::handleRequest
+      Role: !GetAtt ReadCristinAuthenticationRole.Arn
```
torarnet, line 482: You dont need this for an open endpoint

template.yaml
```diff
@@ -473,6 +473,36 @@ Resources:
...
+            Path: /funding-sources
+            Method: get
+
+  NvaCristinFetchFundingSourceFunction:
+    Description: Function for fetching a funding source by identifier
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: cristin-funding
+      Handler: no.unit.nva.cristin.funding.sources.fetch.FetchFundingSourceHandler::handleRequest
+      Role: !GetAtt ReadCristinAuthenticationRole.Arn
```
torarnet, line 497: Same

template.yaml
torarnet: It is what allows function to add auth header when doing authorized requests to cristin

torarnet (dismissed):
Approved with minor comments

cristin-funding/src/test/resources/success.json
ketilaa: I don't need them all. I will reduce it.

template.yaml
ketilaa: I will delete.

template.yaml
ketilaa: I will delete

PR #243, opened 2022-12-20
Np 25966 biobank module

tkardu:
Creating module outline
CHecking gradle
No Karate tests on funding too?


tkardu:
Not sure what is expected to be part of module task else?

settings.gradle
```diff
@@ -5,5 +5,11 @@ include 'cristin-project'
 include 'cristin-organization'
 include 'cristin-person'
 include 'cristin-funding'
+include 'cristin-biobank'
 include 'cristin-testing'
+include 'cristin-biobank'
```
ketilaa, line 10: I think you need only one entry with "include 'cristin-biobank'. Line 10-14 can be removed I think

cristin-biobank/src/main/src/java/no/unit/nva/model/nva/Biobank.java
```diff
@@ -0,0 +1,46 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.net.URI;
+import java.util.Collections;
+import java.util.Map;
+
+@JsonInclude(NON_NULL)
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class FundingSource {
```
ketilaa, line 14: Class name should be Biobank, not FundingSource :)

ketilaa (commented):
Good start. Check my comment in settings.gradle. You should probably fix that. One small comment on class name for model class. It was not part of the task, so you will probably fix that eventually.

ketilaa (approved):

PR #244, opened 2022-12-21
Np 25966 biobank module

tkardu:


ketilaa:
I see that you are still working on the module branch. You should have merged it to main when it was approved, and created a new one for the model part. But I know getting coverage before you have implemented everything might be hard. You temporarily lower the coverage limit for modules that are under development to be able to merge to main still. Up to you.

cristin-biobank/src/main/src/java/no/unit/nva/model/nva/Biobanks.java
```diff
@@ -0,0 +1,48 @@
...
+    @JsonProperty(BIOBANKS_STRING)
+    private final List<Biobank> biobankList;
+
+    @JsonCreator
+    public Biobanks(@JsonProperty(CONTEXT) URI context,
+                          @JsonProperty(ID) URI id,
+                          @JsonProperty(BIOBANKS_STRING) List<Biobank> biobanks) {
+        this.context = context;
+        this.id = id;
+        this.biobankList = Collections.unmodifiableList(biobanks);
```
ketilaa, line 34: When looking at the json representation, biobanks looks better than biobankList. Put if you use biobankList, be consistent and give the constructor parameter the same name.

ketilaa (changes requested):
Looking good. Still fields missing that need to be added :)

cristin-commons/src/main/java/no/unit/nva/utils/DateInfo.java
```diff
@@ -1,4 +1,4 @@
-package no.unit.nva.cristin.projects.model.nva;
+package no.unit.nva.utils;
```
torarnet, line 1: This is a model class, not a util class. I think a better place would be to put DateInfo in model package, where Organization.java is. The module seems right. Things we need for more than one module should be in commons module, as you have done here

cristin-biobank/src/main/java/no/unit/nva/biobank/common/MappingUtils.java
```diff
@@ -0,0 +1,75 @@
...
+                getBiobankMaterialList(cristinBiobank.getCristinBiobankMaterials())
+                );
+    }
+
+    private static URI getBiobankUri(String cristinBiobankId, String domainName, String basePath) {
+        return DomainUriUtils.getBiobankUri(domainName, basePath, cristinBiobankId);
+    }
+
+    private static List<BiobankMaterial> getBiobankMaterialList(List<CristinBiobankMaterial> cristinListOfMaterials) {
+        List nvaListMaterials = new ArrayList<>();
```
torarnet, line 71: Is there i reason not to use typed list? But its not a big issue

cristin-biobank/src/main/java/no/unit/nva/biobank/common/MappingUtils.java
```diff
@@ -0,0 +1,75 @@
...
+                cristinBiobank.getCristinCoordinatinInstitution().getCristinInstitution().getCristinInstitutionId(),
+                INSTITUTIONS_PATH_ELEMENT);
+
+        var unitId = DomainUriUtils.getBiobankParamUri(domainName, basePath,
+                cristinBiobank.getCristinCoordinatinInstitution().getCristinUnit().getCristinUnitId(),
+                UNITS_PATH_ELEMENT);
+
+        var personId = DomainUriUtils.getBiobankParamUri(domainName, basePath,
+                cristinBiobank.getCristinBiobankCoordinator().getCristinPersonIdentifier(),
+                PERSONS_PATH_ELEMENT);
```
torarnet, line 48: Are these null safe? Do we now that identifier from model is always non null?

cristin-biobank/src/main/java/no/unit/nva/biobank/common/MappingUtils.java
```diff
@@ -0,0 +1,75 @@
...
+        return new Biobank(id, cristinBiobank.getCristinBiobankId(),
+                cristinBiobank.getCristinBiobankType(), cristinBiobank.getName(),
+                cristinBiobank.getCristinBiobankLanguage(), cristinBiobank.getCristinBiobankStartDate(),
+                cristinBiobank.getCristinBiobankStoreUntilDate(), cristinBiobank.getCristinBiobankStatus(),
+                new TimeStampFromSource(cristinBiobank.getCristinBiobankCreated()),
+                new TimeStampFromSource(cristinBiobank.getCristinBiobankLastModified()),
+                institutionId,
+                unitId,
+                personId,
+                cristinBiobank.getCristinBiobankAssocProject().getCristinProjectId(),
```
torarnet, line 59: null safe?

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinAssocProjectForBiobank.java
```diff
@@ -0,0 +1,48 @@
...
+    private final Map<String, String> title;
+
+
+    /**
+     * Constructor
+     * @param cristinProjectId - URI
+     * @param cristinProjectIdentificator - code
+     * @param title - name of the project
+     */
+    public CristinAssocProjectForBiobank(URI cristinProjectId,
```
torarnet, line 29: We might want to consider for later annotating with JsonCreator and adding the JsonProperty here as well, but we dont need to do that yet until the need arrives.

Also for later, consider creating the toString, equals and hashcode if we need them

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinBiobank.java
```diff
@@ -0,0 +1,189 @@
...
+
+@SuppressWarnings("PMD.ExcessiveParameterList")
+public final class CristinBiobank {
+    //Here be more params
+    private static final String CRISTIN_BIOBANK_ID = "cristin_biobank_id";
+    private static final String CRISTIN_NAME_FIELD = "name";
+
+    private static final String CRISTIN_BIOBANK_COORDINATING_INSTITUTION = "institution";
+
+    private static final String CRISTIN_BIOBANK_LAST_MODIFIED = "last_modified";
```
torarnet, line 18: This is also a good way to define name for a constant. The important thing when giving names is we understand easily which constant map to each field.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinBiobank.java
```diff
@@ -0,0 +1,189 @@
...
+
+    @JsonProperty(CRISTIN_BIOBANK_TYPE)
+    private final String cristinBiobankType;
+    @JsonProperty(CRISTIN_NAME_FIELD)
+    private final Map<String, String> name;
+    @JsonProperty(CRISTIN_BIOBANK_MAIN_LANGUAGE_KEY)
+    private final String cristinBiobankLanguage;
+
+    @JsonProperty(CRISTIN_BIOBANK_START_DATE)
+    private final Instant cristinBiobankStartDate;
```
torarnet, line 42: We already know that the field is a part of the class CristinBiobank, so I do not think it is necessary to prefix all field names in class CristinBiobank with cristinBiobank. Maybe i need a second opinion

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinCoordinatinInstitution.java
```diff
@@ -0,0 +1,30 @@
+package no.unit.nva.biobank.model.cristin;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import no.unit.nva.cristin.model.CristinInstitution;
+import no.unit.nva.cristin.model.CristinUnit;
+
+public class CristinCoordinatinInstitution {
```
torarnet, line 7: What about this:
https://github.com/BIBSYSDEV/nva-cristin-service/blob/main/cristin-commons/src/main/java/no/unit/nva/cristin/model/CristinOrganization.java
?

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinExternalSourcesBiobank.java
```diff
@@ -0,0 +1,28 @@
+package no.unit.nva.biobank.model.cristin;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class CristinExternalSourcesBiobank {
```
torarnet, line 5: https://github.com/BIBSYSDEV/nva-cristin-service/blob/main/cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/CristinExternalSource.java

torarnet: You can either move that one to commons, or, if you feel there is better to have a duplicate in local module, maybe we should keep it as is

torarnet (commented):
Some little things here and there. We can discuss tomorrow. I might need a second opinion if you can request review from more than me

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinAssocProjectForBiobank.java
torarnet: Is also relevant other places, but not very important yet

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinCoordinatinInstitution.java
tkardu: Originally I was planning somehow to use it from commons,
But then Ketil said that in NVA model biobank we just need URI of the organization and unit + it is not all information about organization which is filled. 
Maybe it is more reasonable to use organization still, true. 
Need second opinion as you said.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinBiobank.java
tkardu: I did mostly for myself and future code exploration. We do know it, but when someone tries to navigate the code it makes it way easier to have it prefixed to avoid confusion?

brinxmat (changes requested):
See other reviewer's comments

cristin-biobank/src/main/java/no/unit/nva/biobank/common/MappingUtils.java
ketilaa: There is no good reason to use untyped List is most cases. Should be changed.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinBiobank.java
ketilaa: Well, class name is CristinBiobank. It is much easier to have a field named startDate(). All the extra chatter prefixing all fields with cristinBiobank is just noise in my opinion.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinCoordinatinInstitution.java
ketilaa: You can use CristinOrganization and populate only id (uri) too.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinExternalSourcesBiobank.java
ketilaa: Why keep duplicates? If it is already in commons, use it.

cristin-biobank/src/main/java/no/unit/nva/biobank/common/MappingUtils.java
tkardu: I have just fixed that.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinExternalSourcesBiobank.java
tkardu: It is not in commons. I will move it there.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinBiobank.java
```diff
@@ -0,0 +1,195 @@
...
+     * @param coordinator - Id of coordinator (person URI)
+     * @param assocProject - Id of the project (URI)
+     * @param externalSources - source short name and reference id
+     * @param approvals - approval parameters
+     * @param materials - material code and description
+     */
+    public CristinBiobank(@JsonProperty(CRISTIN_BIOBANK_ID) String cristinBiobankId,
+                          @JsonProperty(CRISTIN_BIOBANK_TYPE) String type,
+                          @JsonProperty(CRISTIN_BIOBANK_MAIN_LANGUAGE_KEY) String cristinBiobankLanguage,
+                          @JsonProperty(CRISTIN_NAME_FIELD) Map<String, String> name,
```
torarnet, line 101: You should annotate every parameter, not just the first 4

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinBiobank.java
```diff
@@ -0,0 +1,195 @@
...
+                          String status,
+                          CristinTimeStampFromSource created,
+                          CristinTimeStampFromSource cristinBiobankLastModified,
+                          CristinOrganization coordinatinInstitution,
+                          CristinCoordinator coordinator,
+                          CristinAssocProjectForBiobank assocProject,
+                          CristinExternalSource externalSources,
+                          CristinBiobankApprovals approvals,
+                          List<CristinBiobankMaterial> materials) {
+        this.biobankId = cristinBiobankId;
```
torarnet, line 113: use parameter names matching field name, example:
`this.biobankId = biobankId`

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinAssocProjectForBiobank.java
```diff
@@ -0,0 +1,48 @@
...
+import java.net.URI;
+import java.util.Map;
+
+public class CristinAssocProjectForBiobank {
+
+    public static final String CRISTIN_PROJECT_IDENTIFICATOR = "cristin_project_id";
+    public static final String CRISTIN_PROJECT_ID = "url";
+    public static final String CRISTIN_PROJECT_TITLE = "title";
+
+    @JsonProperty(CRISTIN_PROJECT_ID)
```
torarnet, line 14: This points to string value "url" and is hard to understand. Use better names describing the constants.

```
public static final String CRISTIN_PROJECT_ID = "cristin_project_id";
public static final String CRISTIN_PROJECT_URL = "url";
public static final String CRISTIN_PROJECT_TITLE = "title";
```

Or something similar. See earlier comment

cristin-biobank/src/main/java/no/unit/nva/biobank/model/nva/ExternalSourcesBiobank.java
```diff
@@ -0,0 +1,35 @@
+package no.unit.nva.biobank.model.nva;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import no.unit.nva.cristin.model.CristinExternalSource;
+
+public class ExternalSourcesBiobank {
+
+
+    public static final String EXTERNAL_SOURCE_SHORT_NAME = "sourceShortName";
+    public static final String EXTERNAL_SOURCE_REDERENCE_ID = "sourceReferenceId";
```
torarnet, line 10: Ask other reviewer if we should use these names for the data model

torarnet (commented):
See comments

cristin-biobank/src/main/java/no/unit/nva/biobank/common/DomainUriUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Generating URIs for fields of the Biobank, considering their specific path.
+     * @param domainName - env
+     * @param basePath - path
+     * @param identifier - code
+     * @param pathElement -path element for particular parameter
+     */
+    public static URI getBiobankParamUri(String domainName, String basePath, String identifier, String pathElement) {
+        var urlEncodedIdentifier = URLEncoder.encode(identifier, StandardCharsets.UTF_8);
```
ketilaa, line 44: Url encoding the cristinBiobankId is not necessary as it can only contain digits.

cristin-biobank/src/main/java/no/unit/nva/biobank/common/MappingUtils.java
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * @param cristinBiobank - cristin model
+     * @param domainName     - env
+     * @param basePath       - path
+     */
+    public static Biobank cristinModelToNvaModel(CristinBiobank cristinBiobank,
+                                                 String domainName,
+                                                 String basePath
+    ) {
+        var id = getBiobankUri(cristinBiobank.getBiobankId(), domainName, basePath);
```
ketilaa, line 42: You have some typos here I think. Is it not supposed to be `coordinatingInstitution`? Same applies to variable and getter in `CristinBiobank`.

cristin-biobank/src/main/java/no/unit/nva/biobank/common/DomainUriUtils.java
```diff
@@ -0,0 +1,59 @@
...
+        return getBaseUriWrapper(domainName, basePath).addChild(BIOBANK_PATH_ELEMENT).getUri();
+    }
+
+    /**
+     * Building up URI ID for Biobank.
+     * @param domainName - env
+     * @param basePath - path
+     * @param identifier - code
+     */
+    public static URI getBiobankUri(String domainName, String basePath, String identifier) {
```
ketilaa, line 28: Just wondering. Why do you need both `getBiobankUri()` and `getBiobankParamUri()`? Will the path to a biobank be different than `BIOBANK_PATH_ELEMENT` in any cases?

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinBiobank.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @JsonProperty(CRISTIN_BIOBANK_STATUS)
+    private final String status;
+    @JsonProperty(CRISTIN_BIOBANK_CREATED)
+    private final CristinTimeStampFromSource created;
+    @JsonProperty(CRISTIN_BIOBANK_LAST_MODIFIED)
+    private final CristinTimeStampFromSource lastModified;
+
+    @JsonProperty(CRISTIN_BIOBANK_COORDINATING_INSTITUTION)
+    private final CristinOrganization coordinatinInstitution;
```
ketilaa, line 60: `coordinatingInstitution`

cristin-biobank/src/main/java/no/unit/nva/biobank/model/cristin/CristinCoordinatinInstitution.java
ketilaa, line 7: `CristinCoordinatingInstitution`

cristin-biobank/src/main/java/no/unit/nva/biobank/model/nva/Biobank.java
```diff
@@ -0,0 +1,203 @@
...
+    private static final String STORED_UNTIL_DATE = "storedUntilDate";
+    private static final String MAIN_LANGUAGE_KEY = "mainLanguage";
+    private static final String BIOBANK_TYPE = "biobankType";
+    private static final String COORDINATING_INSTITUTION_ORGANIZATION = "coordinating_institution_org";
+    private static final String COORDINATING_INSTITUTION_UNIT = "coordinating_institution_unit";
+
+    private static final String ID_BIOBANK = "id";
+
+    @JsonProperty(ID_BIOBANK)
+    private final transient URI biobankId;
```
ketilaa, line 40: I think you should use transient only if there is no getter made available.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/nva/Biobank.java
```diff
@@ -0,0 +1,203 @@
...
+                   String mainLanguage, Instant startDate,
+                   Instant storeUntilDate, String status,
+                   TimeStampFromSource created, TimeStampFromSource lastModified,
+                   URI coordinatinInstitutionOrg,
+                   URI coordinatinInstitutionUnit, URI biobankCoordinator,
+                   URI assocProject, ExternalSourcesBiobank externalSources,
+                   BiobankApprovals approvals, List<BiobankMaterial> biobankMaterials) {
+        this.biobankId = biobankId;
+        this.biobankIdentifier = biobankIdentifier;
+        this.biobankType = type;
```
ketilaa, line 119: Be consistent in naming. `type` vs `biobankType`

cristin-biobank/src/main/java/no/unit/nva/biobank/model/nva/ExternalSourcesBiobank.java
ketilaa: I would have called this class `ExternalSource`. And the fields `source` and `reference`. Is it a similar model elsewhere in the Cristin API?

cristin-biobank/src/main/java/no/unit/nva/biobank/model/nva/TimeStampFromSource.java
```diff
@@ -0,0 +1,38 @@
+package no.unit.nva.biobank.model.nva;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import no.unit.nva.biobank.model.cristin.CristinTimeStampFromSource;
+
+import java.time.Instant;
+
+public class TimeStampFromSource {
```
ketilaa, line 8: You call the class `TimeStampFromSource`, so fields can just be `timeStamp` and `source`. I would not have use camel-case for timeStamp, just timestamp.

cristin-biobank/src/main/java/no/unit/nva/biobank/model/nva/BiobankMaterial.java
```diff
@@ -0,0 +1,33 @@
+package no.unit.nva.biobank.model.nva;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import no.unit.nva.biobank.model.cristin.CristinBiobankMaterial;
+
+public class BiobankMaterial {
+    public static final String MATERIAL_CODE = "materialCode";
+    public static final String OTHER_MATERIAL_DESC = "otherMaterialDescription";
```
ketilaa, line 8: @torarnet I think description would be enough here, or?

cristin-biobank/src/main/java/no/unit/nva/biobank/model/nva/BiobankMaterial.java
```diff
@@ -0,0 +1,33 @@
...
+    @JsonProperty(OTHER_MATERIAL_DESC)
+    private final String otherMaterialDescription;
+
+    public BiobankMaterial(String materialCode,
+                                  String otherMaterialDescription) {
+        this.materialCode = materialCode;
+        this.otherMaterialDescription = otherMaterialDescription;
+    }
+
+    public BiobankMaterial(CristinBiobankMaterial cristinMaterial) {
```
ketilaa, line 21: While this is ok, I would rather have created a static method for creating an instance of BiobankMaterial from a different object:

`
public static Biobank fromCristinBiobank(CristinBiobank cristinBiobank) {...}
`

Or extract this to some mapping utils if there is no reason why the classes should know about each other.

ketilaa (changes requested):

cristin-biobank/src/main/java/no/unit/nva/biobank/model/nva/BiobankMaterial.java
torarnet: We should ask @brinxmat about domain model.

Also, here is a class for ExternalSource, found in project module:
https://github.com/BIBSYSDEV/nva-cristin-service/blob/main/cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/ExternalSource.java

ketilaa: Looks like you can reuse ExternalSource from project module. Move it to commons.

ketilaa: Yes, it was you that commented on ExternalSource in BiobankMaterial comment :)

torarnet: I was not able to write a comment in the other comment, had to write a comment here instead. dont know why

PR #253, opened 2023-01-20
Np 16247 health fields to enum

torarnet:


cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/HealthProjectType.java
```diff
@@ -0,0 +1,75 @@
...
+
+import static java.lang.String.format;
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.joining;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.util.Optional;
+import nva.commons.apigateway.exceptions.BadRequestException;
+
+public enum HealthProjectType {
```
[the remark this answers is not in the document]
torarnet: Problemet her er at Jackson trenger en default verdi for å koble json og pojo tilfelle ingen match. Jeg prøvde å kaste exception hvis verdien var feil, men jackson overstyrte den med sin egen exception. Vi kan ta en diskusjon på det?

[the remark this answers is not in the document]
torarnet:
```
@JsonCreator
    @SuppressWarnings("unused")
    public static HealthProjectType fromJson(String value) throws BadRequestException {
        return stream(values())
                   .filter(nameType -> nameType.getType().equalsIgnoreCase(value))
                   .findAny()
                   .orElseThrow(() -> new BadRequestException("Bad Request should be the exception"));
                   //.orElse(HealthProjectType.INVALID_VALUE);
    }
```

Gir følgende feil, som vil føre til en 500 Server Error:

`com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of no.unit.nva.cristin.projects.model.nva.HealthProjectType, problem: Bad Request should be the exception`

ketilaa (dismissed):

ketilaa (dismissed):

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/ClinicalTrialPhase.java
```diff
@@ -0,0 +1,58 @@
...
+        return phase;
+    }
+
+    /**
+     * Lookup ClinicalTrialPhase by json. If value not supported, throws exception.
+     */
+    @JsonCreator
+    @SuppressWarnings("unused")
+    public static ClinicalTrialPhase fromJson(String value) {
+        return stream(values())
```
brinxmat, line 35: Since we strictly expect one value, I think it's worth considering:

```
        return stream(values())
                   .filter(nameType -> nameType.getPhase().equalsIgnoreCase(value))
                   .collect(SingletonCollector.tryCollect())
                   .orElseThrow(failure -> new IllegalArgumentException(constructError()));
```

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/ClinicalTrialPhase.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private static String collectValueString() {
+        return stream(values()).map(ClinicalTrialPhase::getPhase).collect(Collectors.joining(DELIMITER));
+    }
+
+    /**
+     * Lookup ClinicalTrialPhase by value. If value not supported, returns null.
+     */
+    public static ClinicalTrialPhase fromValue(String value) {
+        return stream(values())
```
brinxmat, line 53: Similarly

```
        return stream(values())
                   .filter(nameType -> nameType.getPhase().equalsIgnoreCase(value))
                   .collect(SingletonCollector.collectOrElse(null));
```

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/HealthProjectType.java
```diff
@@ -0,0 +1,58 @@
...
+        return type;
+    }
+
+    /**
+     * Lookup HealthProjectType by json. If value not supported, throws exception.
+     */
+    @JsonCreator
+    @SuppressWarnings("unused")
+    public static HealthProjectType fromJson(String value) {
+        return stream(values())
```
brinxmat, line 34: Similarly to other instances

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/HealthProjectType.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private static String collectValueString() {
+        return stream(values()).map(HealthProjectType::getType).collect(Collectors.joining(DELIMITER));
+    }
+
+    /**
+     * Lookup HealthProjectType by value. If value not supported, returns null.
+     */
+    public static HealthProjectType fromValue(String value) {
+        return stream(values())
```
brinxmat, line 52: Ditto

brinxmat (changes requested):

cristin-project/src/main/java/no/unit/nva/cristin/projects/create/CreateCristinProjectHandler.java
```diff
@@ -55,7 +56,7 @@ protected NvaProject processInput(NvaProject input, RequestInfo requestInfo, Con
             throws ApiGatewayException {
 
         verifyRequesterCanEditProjects(requestInfo);
-        var validator = new CreateCristinProjectValidator();
+        Validator<NvaProject> validator = new CreateCristinProjectValidator();
```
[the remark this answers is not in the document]
torarnet: I did it this way so we can see the interface type

torarnet:
@StigNorland Enum.valueOf requires non null. I want to have the option to return null if there is not a matching value

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/HealthProjectData.java
```diff
@@ -15,32 +15,32 @@ public class HealthProjectData implements JsonSerializable {
...
-    private final String clinicalTrialPhase;
+    private final ClinicalTrialPhase clinicalTrialPhase;
 
     /**
      * Nva model for a project having additional health related data.
      */
     @JsonCreator
-    public HealthProjectData(@JsonProperty(TYPE) String type, @JsonProperty(LABEL) Map<String, String> label,
-                             @JsonProperty(CLINICAL_TRIAL_PHASE) String clinicalTrialPhase) {
+    public HealthProjectData(@JsonProperty(TYPE) HealthProjectType type, @JsonProperty(LABEL) Map<String, String> label,
```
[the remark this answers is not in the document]
torarnet: @brinxmat Is this correct. Should i remove JsonProperty from constructors?
@StigNorland I sometimes need to generate new values when writing unit tests, using the constructor. I think i will keep a parameterized constructor for now since I do not have any setters.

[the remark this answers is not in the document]
ketilaa: Immutable classes are a good thing in many ways. Why allow changing class instance fields with setters if they are just DTO/model classes? It only allows for doing bad things if you ask me. This is the reason I mostly user final on instance fields and set them all in the constructor. A builder can help if there too many fields.

cristin-project/src/main/java/no/unit/nva/cristin/projects/create/CreateCristinProjectValidator.java
```diff
@@ -61,6 +52,7 @@ private Set<ValidatedResult> validateProjectInput(NvaProject project) {
         var results = new HashSet<ValidatedResult>();
         if (isNull(project)) {
             results.add(ValidatedResult.Empty);
+            return results;
         }
         if (hasId(project)) {
```
ketilaa, line 57: Should you check if it is missing an id here, or is the method name confusing me?

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/ClinicalTrialPhase.java
```diff
@@ -0,0 +1,58 @@
+package no.unit.nva.cristin.projects.model.nva;
+
+import static java.util.Arrays.stream;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.util.stream.Collectors;
+import nva.commons.core.SingletonCollector;
+
+public enum ClinicalTrialPhase {
```
ketilaa, line 9: Another thing. In our NVA model, do we want to have

`{
  ...
  "phase": "PHASE_ONE",
  ...
}`

or

`{
  ...
  "phase": "1",
  ...
}`

?

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/HealthProjectType.java
```diff
@@ -0,0 +1,58 @@
...
+
+import static java.util.Arrays.stream;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.util.stream.Collectors;
+import nva.commons.core.SingletonCollector;
+
+public enum HealthProjectType {
+
+    DRUGSTUDY("DRUGSTUDY"),
```
ketilaa, line 11: Why having a type in the enum at all? By default Jackson will use the enum name to serialise/deserialise to strings.

cristin-project/src/main/java/no/unit/nva/cristin/projects/create/CreateCristinProjectValidator.java
torarnet: Sending in an identifier is not supported, as that value is something we get from cristin, not the other way

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/ClinicalTrialPhase.java
brinxmat: @StigNorland Missing value throws NPE, which I guess @torarnet wanted to avoid in this instance since the value may be null

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/ClinicalTrialPhase.java
torarnet: I will change the enums in new PR to have to values instead of one. Then we can have one value for NVA and one value for Cristin

ketilaa: Not sure I like the idea to mix all concerns in one enum. We need a simple and clear enum for our NVA model. From Cristin we can just use string if so. Mapping should be in a different package from the NVA model, totally decoupled from the NVA model.

brinxmat (dismissed):

brinxmat (approved):

PR #260, opened 2023-02-03
Np 30622 participant fields added

StigNorland:


torarnet (commented):
Add the new fields to existing create karate test and i will approve. Consider also adding a participant with no identifier only name and see if it goes through

docs/cristin-proxy-swagger.yaml
```diff
@@ -2819,6 +2827,10 @@ components:
             value: 'Kjell Ola'
           - type: 'LastName'
             value: 'Nordmann'
+          - type: 'Email'
```
torarnet, line 2830: Wrong place: Look under `ProjectPostExample`. In my file that is line 2639

docs/cristin-proxy-swagger.yaml
```diff
@@ -1563,6 +1563,14 @@ components:
                     type: string
                     pattern: '.+'
                     description: Last name
+                  email:
```
torarnet, line 1566: See under `ProjectPost` and move this code there

torarnet (commented):
See comments

torarnet (approved):

PR #262, opened 2023-02-06
NP-31724: Added new patch field keywords

torarnet:


torarnet:
Codacy failing because target not currently main

brinxmat (dismissed):
Cannot really tell what a keyword is here, but I have no clue if this is relevant to comment

docs/cristin-proxy-swagger.yaml
```diff
@@ -2067,6 +2067,99 @@ components:
...
+              - NorwegianDirectorateOfHealth
+          status:
+            type: string
+            description: 'Status of the application sent to the approval authority.'
+            enum:
+              - NotApplied
+              - Applied
+              - Approved
+              - Declined
+              - Rejection
```
brinxmat, line 2111: Two things, one of which is likely irrelevant: Rejection is a noun, the rest are adjectives (rejected), less relevant — how is declined different from rejected?

brinxmat (changes requested):

docs/cristin-proxy-swagger.yaml
torarnet: I dont know what the difference is, but it might have to do with an application is not formatted correctly, while the other is formatted correctly but not approved. If you want I can ask someone in Cristin.

By the way, this is not the scope of this PR, but I can inquire if you want

brinxmat (approved):

PR #263, opened 2023-02-13
Np 16256 patch project description fields

torarnet:


brinxmat (approved):

docs/cristin-proxy-swagger.yaml
```diff
@@ -1945,6 +1945,14 @@ components:
           items:
             type: string
             description: URI or numeric value for the identifier of the related project
+        academicSummary:
+          $ref: '#/components/schemas/Languages'
```
ketilaa, line 1949: Seems like Language might benefit from a more generic name in the schema ;)

ketilaa (approved):

docs/cristin-proxy-swagger.yaml
ketilaa: Like `LocalizedText`.

PR #268, opened 2023-02-17
np-35789: add header to avoid being blocked by uio spamfilter

torarnet:


cristin-commons/src/main/java/no/unit/nva/cristin/model/Constants.java
```diff
@@ -62,6 +62,8 @@ public class Constants {
     public static final String HASHTAG = "#";
     public static final String SLASH_DELIMITER = "/";
     public static final String CRISTIN_INSTITUTION_HEADER = "Cristin-Representing-Institution";
+    public static final String BAD_BOT_BYPASS = "UiO-BadBotBypass";
+    public static final String LET_ME_IN = "LetMeIn";
```
torarnet, line 66: Is it okay putting them here or should we use env vars from github?

brinxmat: I would have used an env, but now you need to change the "secret" since it is in git

brinxmat (changes requested):

PR #273, opened 2023-03-07
Np 33003 change model for funding

torarnet:


cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/CristinProjectBuilder.java
```diff
@@ -215,4 +218,17 @@ private static CristinExternalSource toCristinExternalSource(ExternalSource exte
...
     }
+
+    /**
+     * Extracts funding source code from URI giving a valid Cristin source code.
+     */
+    public static String extractFundingSourceCode(URI source) {
+        if (isNull(source)) {
+            return null;
+        }
+        var sourceAsText = source.toString();
```
ketilaa, line 229: You can also just wrap the URI in UriWrapper get call getLastPathElement().

ketilaa (approved):

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/CristinProjectBuilder.java
torarnet: I was thinking about http entities like slash, but if it is already an url it might already be urlencoded

torarnet: This supports both uri format and code in string format

ketilaa (approved):

PR #296, opened 2023-05-15
Np 41723 improve org api versioning and abstraction

torarnet:
Half of the code lines are json testdata. So effective changes about 400 lines. This is just the start of the feature, but will do it in several PR to minimize cognitive load on reviewers


cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinOrgApiClientVersion2.java
```diff
@@ -0,0 +1,66 @@
...
+        return new SearchResponse<Organization>(idUri)
+                   .withHits(organizations)
+                   .withSize(getCount(response, organizations));
+    }
+
+    private List<Organization> getOrganizations(HttpResponse<String> response) throws BadGatewayException {
+        var units = asList(getDeserializedResponse(response, UnitDto[].class));
+
+        return units.stream()
+                   //.parallel()
```
ketilaa, line 60: Remove?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/DefaultOrgQueryClientProvider.java
```diff
@@ -0,0 +1,42 @@
...
+            return getVersionTwo();
+        } else {
+            logger.info(CLIENT_WANTS_VERSION_OF_THE_API_CLIENT, VERSION_ONE);
+            return getVersionOne();
+        }
+    }
+
+    private boolean clientRequestsVersionTwo(RequestInfo requestInfo) {
+        return requestInfo
+                   .getQueryParameterOpt(VERSION)
```
ketilaa, line 30: Do we want to control versions through query params, or rather use Accept header? E.g. `Accept: application/v2+json`

ketilaa (commented):
Some comments for now...

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinOrgApiClientVersion2.java
torarnet: I was thinking of keeping in case we need it later. Then remove when I know if we need it or not. I can remove it for now and add it later

cristin-organization/src/main/java/no/unit/nva/cristin/organization/DefaultOrgQueryClientProvider.java
torarnet: I can change to using accept header, but is your example a valid example or do we need to discuss how we should format the accept header?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/dto/UnitDto.java
```diff
@@ -0,0 +1,80 @@
...
+
+import static no.unit.nva.cristin.common.Utils.nonEmptyOrDefault;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+import no.unit.nva.commons.json.JsonSerializable;
+
+public class UnitDto implements JsonSerializable {
```
StigNorland, line 11: This is easier to read, than verbose annotation with string constants.

StigNorland (approved):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinOrgApiClientVersion2.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+import no.unit.nva.cristin.model.SearchResponse;
+import no.unit.nva.cristin.organization.dto.OrganizationFromUnitMapper;
+import no.unit.nva.cristin.organization.dto.UnitDto;
+import no.unit.nva.model.Organization;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+import nva.commons.apigateway.exceptions.BadGatewayException;
+
+public class CristinOrgApiClientVersion2 extends CristinOrganizationApiClient implements IQueryApiClient<Organization> {
```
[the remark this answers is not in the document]
torarnet: The whole point of the new version is to get rid of recursion when fetching from upstream. I will try to keep the model as equal as possible, but some changes might be needed

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinOrgApiClientVersion2.java
```diff
@@ -0,0 +1,66 @@
+package no.unit.nva.cristin.organization;
```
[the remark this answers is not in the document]
torarnet: have put models in own package. will probably do the same with clients when i restructure the package tree. Since it now has more classes there is a need to rework the package tree

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/IClientProvider.java
```diff
@@ -0,0 +1,7 @@
+package no.unit.nva.cristin.common.client;
+
+public interface IClientProvider<T> {
```
brinxmat, line 3: Let's not write C#. I don't think marking interfaces as interfaces is wise — in fact, it is something we explicitly don't want since this is code that is designed to express the business need, not the implementation specifics (I consider the fact that it is an interface to be one)

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/IQueryApiClient.java
```diff
@@ -0,0 +1,10 @@
+package no.unit.nva.cristin.common.client;
+
+import no.unit.nva.cristin.model.SearchResponse;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+
+public interface IQueryApiClient<T, R> {
```
brinxmat, line 6: Ditto

cristin-commons/src/main/java/no/unit/nva/cristin/model/JsonPropertyNames.java
```diff
@@ -70,5 +70,6 @@ public class JsonPropertyNames {
     public static final String UNIT = "unit";
     public static final String USER = "user";
     public static final String LABELS = "labels";
+    public static final String NEAREST_PART_OF = "nearestPartOf";
```
brinxmat, line 73: What is this property designed to express?

cristin-commons/src/main/java/no/unit/nva/utils/VersioningUtils.java
```diff
@@ -15,4 +18,15 @@ public static String extractVersion(String headerValue) {
         return contentType.getParameter(VERSION);
     }
 
+    /**
+     * Extract the version field value if present from header with key directly from RequestInfo or else null.
+     */
+    public static String extractVersionFromRequestInfo(RequestInfo requestInfo, String headerKey) {
```
brinxmat, line 24: Can `headerKey` be inlined in the method since it is constant and part of the method name?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinOrgApiClientVersion2.java
```diff
@@ -0,0 +1,74 @@
...
+    public CristinOrgApiClientVersion2() {
+        this(defaultHttpClient());
+    }
+
+    public CristinOrgApiClientVersion2(HttpClient client) {
+        super(client);
+    }
+
+    /**
+     * Fetch Organizations matching given query criteria using version 2 code.
```
brinxmat, line 40: The version name is not v2, though

brinxmat (changes requested):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinOrgApiClientVersion2.java
StigNorland: What should it be called?

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/IClientProvider.java
StigNorland: I am aware that this is C# naming, but I never have had a good explanation why it is frown upon in Java.
Do you have one?

cristin-commons/src/main/java/no/unit/nva/utils/VersioningUtils.java
torarnet: If we want a versioned post or patch we could pass header Content-Type instead of Accept, so i decided to pass the header key as a parameter

cristin-commons/src/main/java/no/unit/nva/cristin/model/JsonPropertyNames.java
torarnet: Parent_unit object as opposed to list parent_units. When parsing the list into partOf from the list the object nearestPartOf might be redundant

cristin-organization/src/main/java/no/unit/nva/cristin/organization/CristinOrgApiClientVersion2.java
torarnet: And what should we call the packages containing the versions? Now it is "version20230526". Numeric only gives error

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/IQueryApiClient.java
```diff
@@ -0,0 +1,10 @@
...
+
+    SearchResponse<R> executeQuery(T params) throws ApiGatewayException;
```
ketilaa, line 8: Dersom T alltid vil være `Map<String, String>` slik det ser ut for meg, trenger den ikke å være generic.

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/IClientProvider.java
```diff
@@ -0,0 +1,7 @@
...
+
```
ketilaa, line 4: Føler at vi bør kunne si noe mer om hva T er her. Feks:
`IClientProvider<T extends IQueryApiClient<Map<String, String>, ?>>`

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/IQueryApiClient.java
ketilaa: Og dersom vi ikke har javadoc, bruk gode navn på generics, feks:
`IQueryApiClient<T, DTO>`

torarnet: Ved utvidelse av bruken av denne kan vi feks måtte sende inn et objekt av type QueryProject, så jeg tenker det kan være lurt å ha den generisk

torarnet: https://pmd.sourceforge.io/pmd-6.20.0/pmd_rules_java_codestyle.html#genericsnaming

cristin-organization/src/main/java/no/unit/nva/cristin/organization/DefaultOrgQueryClientProvider.java
```diff
@@ -0,0 +1,37 @@
...
+        if (VERSION_2023_05_26.equals(apiVersion)) {
+            logger.info(CLIENT_WANTS_VERSION_OF_THE_API_CLIENT, VERSION_2023_05_26);
+            return getVersion20230526();
+        }
+        logger.info(CLIENT_WANTS_VERSION_OF_THE_API_CLIENT, VERSION_ONE);
+        return getVersionOne();
+    }
+
+    public QueryApiClient<Map<String, String>, Organization> getVersionOne() {
+        return new CristinOrganizationApiClient();
```
StigNorland, line 31: Thought for reflection...
I know that in this project we extract one liners into functions, but does this really improve readability?

StigNorland (approved):
Have one comment, but no change requests.

ketilaa (approved):
Some comments. I don't feel we need to change anything right now. But I feel this versioning design is a bit CristinProxy specific and need some redesign from the bottom up to be something we can use in all APIs. But we can look at that at a later point.

cristin-organization/src/main/java/no/unit/nva/cristin/organization/DefaultOrgQueryClientProvider.java
torarnet: We can mock the wrapper methods when testing...

brinxmat (approved):

PR #297, opened 2023-05-15
np-41723: Creating an abstraction for api client in fetch organization

torarnet:
removed redundant tests
updated test code


cristin-organization/src/main/java/no/unit/nva/cristin/organization/IFetchApiClient.java
```diff
@@ -0,0 +1,9 @@
+package no.unit.nva.cristin.organization;
+
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+
+public interface IFetchApiClient<T, R> {
+
+    R executeFetch(T params) throws ApiGatewayException;
```
torarnet, line 7: Should i rather use a Map as param type?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/DefaultOrgFetchProvider.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Map;
+import no.unit.nva.model.Organization;
+import nva.commons.apigateway.RequestInfo;
+
+public class DefaultOrgFetchProvider implements IClientProvider
+                                                    <RequestInfo, IFetchApiClient<Map<String, String>, Organization>> {
+
+    @Override
+    public IFetchApiClient<Map<String, String>, Organization> getClient(RequestInfo params) {
```
torarnet, line 11: Should i assume we always use RequestInfo as param so we dont have to specify type whenever we use interface IClientProvider?

cristin-organization/src/test/java/no/unit/nva/cristin/organization/FetchCristinOrganizationHandlerTest.java
```diff
@@ -141,28 +140,6 @@ void shouldReturnBadRequestResponseOnMissingPathParam() throws IOException {
         assertThat(actualDetail, containsString(ERROR_MESSAGE_INVALID_PATH_PARAMETER_FOR_ID_FOUR_NUMBERS));
     }
 
-    @Test
-    void shouldReturnInternalServerErrorResponseOnUnexpectedException() throws IOException {
```
torarnet, line 145: this is generated by the framework. dont see any reason to explicit test it again

cristin-organization/src/main/java/no/unit/nva/cristin/organization/DefaultOrgFetchProvider.java
ketilaa: It is not used for now it seems. Personally, I would avoid letting RequestInfo leak into any code but the handler itself.

cristin-organization/src/main/java/no/unit/nva/cristin/organization/IFetchApiClient.java
ketilaa: If it always is a map, use a map.

cristin-organization/src/test/java/no/unit/nva/cristin/organization/FetchCristinOrganizationHandlerTest.java
```diff
@@ -43,54 +40,56 @@
...
 
     @BeforeEach
     void setUp() {
         context = mock(Context.class);
-        mockHttpClient = mock(HttpClient.class);
+        HttpClient mockHttpClient = mock(HttpClient.class);
         cristinApiClient = new CristinOrganizationApiClient(mockHttpClient);
+        cristinApiClient = spy(cristinApiClient);
+        doReturn(Try.of(new HttpResponseFaker(EMPTY_JSON)))
+            .when(cristinApiClient).sendRequestMultipleTimes(any());
```
ketilaa, line 77: I don't like the method name `sendRequestMultipleTimes()`. Either use just `sendRequest()`, or if it is important that it does retry, consider `sendRequestWithRetry()`. `sendRequestMultipleTimes()` indicates to me that it always send the request multiple times.

ketilaa (approved):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/IFetchApiClient.java
StigNorland: T is fine, you are then free to choose implementation later.

cristin-organization/src/test/java/no/unit/nva/cristin/organization/FetchCristinOrganizationHandlerTest.java
StigNorland: sendRequestWithRetry

cristin-organization/src/main/java/no/unit/nva/cristin/organization/DefaultOrgFetchProvider.java
StigNorland: Here is where you define your T,  Alternatively you could set it to String apiVersion, as that is what you need to know.

StigNorland (changes requested):
Vudér om du vil bytte til api-versjon av typen Streng...
Ellers veldig elegant.

cristin-organization/src/test/java/no/unit/nva/cristin/organization/FetchCristinOrganizationHandlerTest.java
```diff
@@ -43,54 +40,56 @@
...
+    private DefaultOrgFetchProvider clientProvider;
     private ByteArrayOutputStream output;
     private Context context;
-    private HttpClient mockHttpClient;
 
     @BeforeEach
     void setUp() {
         context = mock(Context.class);
-        mockHttpClient = mock(HttpClient.class);
+        HttpClient mockHttpClient = mock(HttpClient.class);
```
brinxmat, line 73: var

brinxmat (changes requested):

cristin-organization/src/test/java/no/unit/nva/cristin/organization/FetchCristinOrganizationHandlerTest.java
torarnet: This is just for compatibility with existing code. in the new version i will use another method

StigNorland (changes requested):
One comment, if you have an answer, I have nothing more to add 👍

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPerson.java
```diff
@@ -156,6 +157,13 @@ public void setDetailedAffiliations(List<CristinPersonEmployment> detailedAffili
         this.detailedAffiliations = detailedAffiliations;
     }
 
+    public Boolean getIdentifiedCristinPerson() {
+        return identifiedCristinPerson;
+    }
+
+    public void setIdentifiedCristinPerson(Boolean identifiedCristinPerson) {
```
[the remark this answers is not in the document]
StigNorland: https://unit.atlassian.net/browse/NP-2729?atlOrigin=eyJpIjoiMWVjNzMzZGEyM2VmNGFmZjg5M2I1ZWY0MTc4ZTUzZGEiLCJwIjoiaiJ9

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/HandlerUtil.java
```diff
@@ -0,0 +1,35 @@
...
+            return requestInfo.getQueryParameters().containsKey(DEPTH)
+                       ? requestInfo.getQueryParameter(DEPTH)
+                       : TOP;
+        } else {
+            throw new BadRequestException(ErrorMessages.ERROR_MESSAGE_DEPTH_INVALID);
+        }
+    }
+
+    private static boolean isValidDepth(RequestInfo requestInfo) throws BadRequestException {
+        return !requestInfo.getQueryParameters().containsKey(DEPTH)
```
StigNorland, line 30:
```
   var depth = requestInfo.getQueryParameterOpt(DEPTH).orElse(TOP);
   validateDepth(depth);
   return depth;
...
void validateDepth(depth) {
   if (!Set.of(TOP, FULL, NONE).contains(depth) {
      throw new BadRequestException(ErrorMessages.ERROR_MESSAGE_DEPTH_INVALID);
   }
}
```
Simpler logic?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/version20230526/CristinOrgApiClient20230526.java
```diff
@@ -0,0 +1,140 @@
...
+import no.unit.nva.cristin.common.client.QueryApiClient;
+import no.unit.nva.cristin.model.SearchResponse;
+import no.unit.nva.cristin.organization.dto.version20230526.mapper.OrganizationFromUnitMapper;
+import no.unit.nva.cristin.organization.dto.version20230526.UnitDto;
+import no.unit.nva.model.Organization;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+import nva.commons.apigateway.exceptions.BadGatewayException;
+import nva.commons.core.paths.UriWrapper;
+
+public class CristinOrgApiClient20230526 extends ApiClient
```
StigNorland, line 34: class is easy to read 👍

cristin-organization/src/main/java/no/unit/nva/cristin/organization/fetch/DefaultOrgFetchClientProvider.java
```diff
@@ -0,0 +1,37 @@
...
+import org.slf4j.LoggerFactory;
+
+public class DefaultOrgFetchClientProvider
+    implements ClientProvider<FetchApiClient<Map<String, String>, Organization>> {
+
+    private static final Logger logger = LoggerFactory.getLogger(DefaultOrgFetchClientProvider.class);
+
+    @Override
+    public FetchApiClient<Map<String, String>, Organization> getClient(String apiVersion) {
+        if (VERSION_2023_05_26.equals(apiVersion)) {
```
StigNorland, line 22: What if a client ask for version 2023_06_19, should we return first or last prior version?

cristin-organization/src/test/java/no/unit/nva/cristin/organization/fetch/FetchCristinOrganizationHandlerTest.java
```diff
@@ -242,6 +224,31 @@ void shouldReturnOrganizationFlatHierarchy() throws IOException {
...
+
+        fetchCristinOrganizationHandler = new FetchCristinOrganizationHandler(clientProvider, new Environment());
+        fetchCristinOrganizationHandler.handleRequest(requestUsingVersionTwo(), output, context);
+
+        var gatewayResponse = GatewayResponse.fromOutputStream(output, Organization.class);
+        var responseBody = gatewayResponse.getBodyObject(Organization.class);
+
+        var pretty = OBJECT_MAPPER.readTree(responseBody.toString()).toPrettyString();
+
+        System.out.println(pretty);
```
StigNorland, line 248: use logger instead

StigNorland (changes requested):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/fetch/DefaultOrgFetchClientProvider.java
torarnet: defaults then to first version

cristin-organization/src/test/java/no/unit/nva/cristin/organization/fetch/FetchCristinOrganizationHandlerTest.java
torarnet: Sorry, work in progress. will fix before asking for new review

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/HandlerUtil.java
torarnet: Seems to work, will change to your example

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/version20230526/CristinOrgApiClient20230526.java
torarnet: i am considering splitting query and fetch into their own separate clients, because i dont think we need shared methods after all. The goal was to either doing a search with a implicit full query for each, but then it might have gotten very slow and verbose. so im thinking now that client does a separate search and then fetches one at a time for their requested search result

StigNorland (dismissed):

cristin-organization/src/test/java/no/unit/nva/cristin/organization/query/QueryCristinOrganizationHandlerTest.java
```diff
@@ -194,6 +199,40 @@ void shouldReturnVersionOneAsDefault() throws Exception {
...
+        var fakeQueryResponseResource = IoUtils.stringFromResources(Path.of(CRISTIN_QUERY_RESPONSE_V_2_JSON));
+        doReturn(new HttpResponseFaker(fakeQueryResponseResource))
+            .when(cristinOrgApiClient20230526).fetchQueryResults(any());
+
+        queryCristinOrganizationHandler = new QueryCristinOrganizationHandler(clientProvider, new Environment());
+        var input =
+            generateValidHandlerRequestWithVersionParam(String.format(ACCEPT_HEADER_EXAMPLE, VERSION_2023_05_26));
+        queryCristinOrganizationHandler.handleRequest(input, output, context);
+
+        var gatewayResponse = GatewayResponse.fromOutputStream(output,
```
StigNorland, line 213: This has nothing to with your code, just something that bothers me, with unneccessary code.
Wouldn't it be nice to skip the second Class ref, as you allready supplied it in the line before ?

StigNorland (approved):
Just an after thought.
We now use createdDate as version, but it will likely be updated manytimes during its lifetime, 
having a date attached to each file might not be helpful in the end.

You already have a `dto/.v20230526/UniDto.java` which makes sense when you just add a new Class, maybe we should make them all like this?

.github/workflows/karate.yml
```diff
@@ -59,5 +59,5 @@ jobs:
       - name: Delete karate test users
         run: BASE_PATH=${{env.BASE_PATH}} ./gradlew deleteTestUsers --info
         # Delete stack after running tests
-      - name: Delete stack after running tests
-        run: sam delete --no-prompts --region eu-west-1 --stack-name ${{env.STACK_NAME}}
+      #- name: Delete stack after running tests
```
torarnet, line 62: Remind me to revert this when ready to merge

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/ClientProvider.java
StigNorland, line 1: First I was like, waf does these do here, in a generic interface, but is not, and it was a good move....

cristin-organization/src/main/java/no/unit/nva/cristin/organization/dto/v20230526/mapper/OrganizationFromUnitMapper.java
```diff
@@ -40,9 +39,10 @@ private Set<Organization> unitsToOrganizations(List<UnitDto> unitDtos) {
     }
 
     private Set<Organization> parentToSetOfOrganizations(UnitDto parent) {
-        var set = new HashSet<Organization>();
-        set.add(this.apply(parent));
-        return set;
+        return Optional.ofNullable(parent)
```
StigNorland, line 42: This is just a fad, but I really like this syntax, even if it is probably slower than the old one. 👍

StigNorland (approved):
How about renaming namespace/folder to /v2023-05-26/ ?
and keep different version of same class with equal names?

It would certainly improve readability, 
but could create confusion as to which class you are working on.

I am unsure what would be best.

cristin-organization/src/main/java/no/unit/nva/cristin/organization/dto/v20230526/mapper/OrganizationFromUnitMapper.java
torarnet: The old syntax gave a list with a element with value null, so i fixed it and rather just returned an empty list, not a list with element null

cristin-organization/src/main/java/no/unit/nva/cristin/organization/query/DefaultOrgQueryClientProvider.java
```diff
@@ -19,12 +18,17 @@
 
     @Override
     public QueryApiClient<Map<String, String>, Organization> getClient(String apiVersion) {
-        if (VERSION_2023_05_26.equals(apiVersion)) {
-            logger.info(CLIENT_WANTS_VERSION_OF_THE_API_CLIENT, VERSION_2023_05_26);
-            return getVersion20230526();
+        switch (nonNull(apiVersion) ? apiVersion : EMPTY_STRING) {
```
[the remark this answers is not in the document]
torarnet: Its a smart move considering we might add more versions later 😃

cristin-organization/src/main/java/no/unit/nva/cristin/organization/dto/v20230526/mapper/OrganizationFromUnitMapper.java
StigNorland: ok, so not just a fad :)

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/v20230526/FetchCristinOrgClient20230526.java
```diff
@@ -0,0 +1,96 @@
...
+     * response should include sub organization tree downwards or just have the upwards parent trail.
+     *
+     * @param params Map containing verified query parameters
+     */
+    @Override
+    public Organization executeFetch(Map<String, String> params) throws ApiGatewayException {
+        var identifier = params.get(IDENTIFIER);
+        var fetchUri = getCristinUri(identifier);
+        var response = fetchGetResult(fetchUri);
+        if (wantsDepth(params)) {
```
brinxmat, line 51: hasDepth?

cristin-organization/src/test/java/no/unit/nva/cristin/organization/fetch/FetchCristinOrganizationHandlerTest.java
```diff
@@ -208,6 +236,48 @@ void shouldReturnOrganizationFlatHierarchy() throws IOException {
         assertThat(actualOrganization.getPartOf(), equalTo(null));
     }
 
+    @Test
+    void shouldNotCallUpstreamOneMoreTimeForFetchingSubunitsWhenSendingParamDepthWithValueNone() throws Exception {
```
brinxmat, line 240: Times

brinxmat (changes requested):

cristin-organization/src/test/java/no/unit/nva/cristin/organization/fetch/FetchCristinOrganizationHandlerTest.java
torarnet: shouldNotCallUpstreamMoreTimesFor...?

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/v20230526/FetchCristinOrgClient20230526.java
torarnet: hasDepth is not entirely correct. We dont know yet that the organization has depth at this stage. But if the client has not specified he does not want depth, we assume wantsDepth == true and then do an extra request for fetching requested organization's subunits. We dont get everything in one request because of how upstream is modelled

brinxmat (approved):

PR #299, opened 2023-05-23
Np 44746 test http version one

torarnet:


cristin-project/src/main/java/no/unit/nva/cristin/projects/update/UpdateCristinProjectApiClient.java
```diff
@@ -19,16 +18,6 @@
 
 public class UpdateCristinProjectApiClient extends PatchApiClient {
 
-    /**
-     * Create a cristin API client with default HTTP client.
-     */
-    public UpdateCristinProjectApiClient() {
```
torarnet, line 25: unused...

cristin-project/src/main/java/no/unit/nva/cristin/projects/fetch/FetchCristinProjectApiClient.java
```diff
@@ -25,10 +24,7 @@ public class FetchCristinProjectApiClient extends CristinProjectApiClient {
      * Creates instance with default Http Client.
      */
     public FetchCristinProjectApiClient() {
-        this(HttpClient.newBuilder()
-                 .followRedirects(HttpClient.Redirect.ALWAYS)
-                 .connectTimeout(Duration.ofSeconds(30))
-                 .build());
+        super();
```
torarnet, line 27: super constructor constructs the client itself

cristin-commons/src/main/java/no/unit/nva/HttpClientProvider.java
```diff
@@ -0,0 +1,17 @@
...
+import java.net.http.HttpClient.Version;
+import java.time.Duration;
+
+public class HttpClientProvider {
+
+    public static HttpClient defaultHttpClient() {
+        return HttpClient.newBuilder()
+                   .followRedirects(HttpClient.Redirect.ALWAYS)
+                   .version(Version.HTTP_1_1)
+                   .connectTimeout(Duration.ofSeconds(30))
```
ketilaa, line 13: Occupying a server thread in Cristin for 30 seconds is a long time. Does it even exceed our lambda timeout? What would happen if we decreased this to e.g. 10 seconds from your experience?

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/CristinAuthenticator.java
```diff
@@ -42,10 +43,11 @@ protected PasswordAuthentication getPasswordAuthentication() {
      */
     public static HttpClient getHttpClient() {
         return HttpClient.newBuilder()
-                .followRedirects(HttpClient.Redirect.ALWAYS)
-                .authenticator(getBasicAuthenticator())
-                .connectTimeout(Duration.ofSeconds(30))
-                .build();
+                   .followRedirects(HttpClient.Redirect.ALWAYS)
```
ketilaa, line 46: Why not use the HttpClientProvider here?

ketilaa (dismissed):

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/CristinAuthenticator.java
ketilaa: Ahh, the authenticator. Did you consider adding support for authenticator in HttpClientProvider?

torarnet: we use `.authenticator` which is not present in the default client method

torarnet: Did not consider adding support in HttpClientProvider since it is only used one place, and that is here.

cristin-commons/src/main/java/no/unit/nva/HttpClientProvider.java
torarnet: Should i change it to 10, maybe 15 seconds before merge or keep as is?

torarnet: I dont think it will be a problem to decrease to 15, but under high load in cristin i have experienced up to 1 minute response times, which gives us a gateway timeout from the load balancer

ketilaa (approved):

PR #301, opened 2023-05-30
Np 44788 add field identified cristin person

torarnet:


docs/cristin-proxy-swagger.yaml
```diff
@@ -2417,6 +2417,9 @@ components:
           description: Employments for this person. Visible to authorized users
           items:
             $ref: '#/components/schemas/PersonEmployment'
+        verified:
+          type: boolean
+          description: If the person is a verified person with a national identity number
```
brinxmat, line 2422: I'm not sure that this will hold over time — perhaps write "Whether we can document the existence of the person"

brinxmat (changes requested):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
```diff
@@ -33,6 +33,7 @@
 import static no.unit.nva.cristin.person.model.nva.JsonPropertyNames.NAMES;
 import static no.unit.nva.cristin.person.model.nva.JsonPropertyNames.NATIONAL_IDENTITY_NUMBER;
 import static no.unit.nva.cristin.person.model.nva.JsonPropertyNames.RESERVED;
+import static no.unit.nva.cristin.person.model.nva.JsonPropertyNames.VERIFIED;
 
 @JacocoGenerated
 @JsonPropertyOrder({CONTEXT, ID, TYPE, IDENTIFIERS, NAMES, CONTACT_DETAILS, IMAGE, AFFILIATIONS})
```
StigNorland, line 39: This is something to consider to add to every DTO

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
```diff
@@ -59,6 +60,8 @@ public class Person implements JsonSerializable {
     private Boolean reserved;
     @JsonProperty(EMPLOYMENTS)
     private Set<Employment> employments;
+    @JsonProperty(VERIFIED)
```
StigNorland, line 63: I am not a fan of this way of defining dto's, but you are following the pattern that was already there,  
please read & ignore.

cristin-person/src/test/java/no/unit/nva/cristin/person/handler/person.feature
```diff
@@ -158,3 +158,12 @@ Feature: API tests for Cristin Person fetch
...
     And match identifiers !contains 'NationalIdentificationNumber'
+
+  Scenario: Get returns supported fields in payload
+    * def samplePersonIdentifier = '515114'
+    Given path '/person/' + samplePersonIdentifier
+    When method GET
+    Then status 200
+    And match response == '#object'
+    And match response.verified == '#present'
+    And match response.verified == true
```
StigNorland, line 169: Question, does response.verified == true  return false if verified is missing or undefined? 
Or does it crash?

docs/cristin-proxy-swagger.yaml
StigNorland: Foreign workers might get new national ID, depending on what kind of ID they are entering our border with,

StigNorland (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
torarnet: Another one on the team said earlier that property order is not important because we do not need it other than that it looks better to read for a human. Just keeping it because it was already there

torarnet: Would love to hear more about your opinion by the way, it might be reasons to use it after all

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
torarnet: Exactly, did not want to change status quo for now, might be room for refactor later

cristin-person/src/test/java/no/unit/nva/cristin/person/handler/person.feature
torarnet: It must be there, if missing it will fail the test. But this is a test person which would not change status verified

brinxmat (approved):

PR #304, opened 2023-06-20
Np 44887 new field external url

torarnet:


cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/NvaProject.java
```diff
@@ -373,6 +376,14 @@ public void setCreator(NvaContributor creator) {
         this.creator = creator;
     }
 
+    public URI getWebPage() {
```
brinxmat, line 379: This may be null, sometimes I wonder if Optional is a good return type to make this clear…possibly not in this case.

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/NvaProjectBuilder.java
```diff
@@ -150,9 +151,14 @@ public NvaProject build() {
                    .withApprovals(extractApprovals(cristinProject.getApprovals()))
                    .withExemptFromPublicDisclosure(cristinProject.getExemptFromPublicDisclosure())
                    .withCreator(extractCreator(cristinProject.getCreator()))
+                   .withWebPage(extractWebPage(cristinProject.getExternalUrl()))
                    .build();
     }
 
+    private URI extractWebPage(String externalUrl) {
```
brinxmat, line 158: Right the optional of getter does make sense

cristin-project/src/test/java/no/unit/nva/cristin/projects/fetch/project.feature
```diff
@@ -167,3 +167,11 @@ Feature: API tests for Cristin Project retrieve and search
     And match response.approvals[0].applicationCode == 'EthicalApproval'
     And match response.approvals[0].identifier == '2017/800'
     And match response.approvals[0].authorityName == '#present'
+
+  Scenario: Fetch returns project with webpage
```
brinxmat, line 171: You might want the non-webpage test too

anetteOlli (dismissed):
Lest igjennom. Fant ikke noe åpenbart å pirke på

cristin-project/src/main/java/no/unit/nva/cristin/projects/update/ProjectPatchValidator.java
```diff
@@ -237,4 +240,15 @@ private void validateExtraLanguages(ObjectNode input) throws BadRequestException
     private BadRequestException notAListOfMapsException() {
         return new BadRequestException(format(NOT_A_VALID_LIST_OF_KEY_VALUE_FIELDS, ALTERNATIVE_TITLES));
     }
+
+    private void validateWebPage(ObjectNode input) throws BadRequestException {
+        if (input.has(WEB_PAGE)) {
```
StigNorland, line 245:
```
        Optional.ofNullable(input.get(WEB_PAGE))
            .map(JsonNode::asText)
            .map(URI::create)
            .orElseThrow(failure -> fieldIsNotUri());
```
If you want to follow the same style as in `CristinProjectBuilder.java`

cristin-project/src/test/java/no/unit/nva/cristin/projects/RandomProjectDataGenerator.java
```diff
@@ -120,6 +121,7 @@ public static NvaProject randomNvaProject() {
                                           .withHealthProjectData(randomHealthProjectData())
                                           .withApprovals(randomApprovals())
                                           .withExemptFromPublicDisclosure(randomBoolean())
+                                          .withWebPage(randomUri())
```
StigNorland, line 124: How about adding a randomUri(withEndpoint) to give us a clue to where a randomUri should have pointed to?

StigNorland (approved):
I added some thoughts that has nothing to do with the code in question, except for use of Optional, which you may concider changing (adding)

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/NvaProject.java
```diff
@@ -105,6 +105,9 @@ public class NvaProject implements JsonSerializable {
     @JsonProperty
     @JsonInclude(NON_NULL)
     private NvaContributor creator;
+    @JsonProperty
+    @JsonInclude(NON_NULL)
```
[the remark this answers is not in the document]
torarnet: So what you are saying is if a property does not have a value, we should display it as null in the json? Then i think we should use JsonInclude(ALWAYS) and not hide fields with value null?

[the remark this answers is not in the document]
StigNorland: That is my opinion (and it is easier to handle for the frontend).

ketilaa (approved):

cristin-project/src/test/java/no/unit/nva/cristin/projects/RandomProjectDataGenerator.java
torarnet: this webpage can be anything, it does not have a common domain

cristin-project/src/main/java/no/unit/nva/cristin/projects/update/ProjectPatchValidator.java
torarnet: input.has and input.get works a little different. Having something to do with the field present or not, and the field being null or not when present. Tests failed when i tried you example

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/NvaProjectBuilder.java
torarnet: no other field in this model class is using optional. if we want to start using it i suggest we do it for all optional fields in a separate PR. we should also discuss one day using field = null or field omitted if there is no value there.

cristin-project/src/main/java/no/unit/nva/cristin/projects/update/ProjectPatchValidator.java
StigNorland: Did you know that ?
Wouldn't that have implications for when we are serializing empty fields to storage.
(we omit them, instead of adding with null)

cristin-project/src/test/java/no/unit/nva/cristin/projects/fetch/project.feature
torarnet: We might need to decide if field should be null or just omitted first, but i can always add a test now and change it later

cristin-project/src/main/java/no/unit/nva/cristin/projects/update/ProjectPatchValidator.java
torarnet: yes i did not it but tested your suggestion just to see if we could simplify. When doing patch requests we have to build the json manually because patch allows us to explicit send a field value as null, so we have to know if the client sends the field at all, sends the field with value null, or field with a regular value

torarnet: Also added a test for checking if field will be sent as null to upstream

StigNorland: Maybe something that the whole team should know about ?

StigNorland (approved):

brinxmat (approved):

PR #305, opened 2023-06-21
Migrate name to labels, adding context-object

brinxmat:
- name is labels
- add context object


torarnet (commented):
Update karate tests with new field name and validate context object is displayed correctly

cristin-funding/src/main/java/no/unit/nva/cristin/funding/sources/model/nva/FundingSource.java
```diff
@@ -21,15 +31,15 @@ public class FundingSource {
     @JsonProperty(IDENTIFIER_FIELD)
     private final String identifier;
     @JsonProperty(NAME_FIELD)
```
torarnet, line 33: you forgot to update name tag

torarnet (changes requested):
You forgot to update field's name tag. Also, you need to update karate tests

ketilaa:
@brinxmat Did you forget about this PR? Is it something we should work to complete?

torarnet:
status @brinxmat ?

PR #323, opened 2023-09-04
Np 44703 allow profile picture deletion

torarnet:


StigNorland (changes requested):
Why should-Allow-Deleting-Profile-Picture-By-Sending-`Empty-Payload` ?
technical issue or ?

In documentation it says that size is increaed to 1mb, but I couldn't see any chenages to the code, is it already implemented?

torarnet:
Empty payload transforms to empty byte[] which then removes an existing picture in upstream. Size of 1mb was done in upstream, not in proxy

PR #324, opened 2023-09-04
Np 45210 unable to fetch profile picture

torarnet:


torarnet:
Dont know if this fixed the problem, but the stack now displays picture data. Added a integration test so we can capture this if it happens again

cristin-person/src/main/java/no/unit/nva/cristin/person/picture/fetch/FetchPictureHandler.java
```diff
@@ -12,12 +12,13 @@
...
 
 public class FetchPictureHandler extends ApiGatewayHandler<Void, Binary> {
+
     private final transient FetchPictureApiClient apiClient;
 
     @SuppressWarnings("unused")
     @JacocoGenerated
     public FetchPictureHandler() {
-        this(new FetchPictureApiClient(CristinAuthenticator.getHttpClient()), new Environment());
+        this(new FetchPictureApiClient(defaultHttpClient()), new Environment());
```
StigNorland, line 21: Maybe we need to go through all repositories and fix this ?

StigNorland (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/picture/fetch/FetchPictureHandler.java
torarnet: No, that is not neccessary. The problem here might have that we used an authenticated client while we ALSO set the authorization header manually.

See line 70 here:
https://github.com/BIBSYSDEV/nva-cristin-service/blob/main/cristin-person/src/main/java/no/unit/nva/cristin/person/picture/fetch/FetchPictureApiClient.java

torarnet: But TBH, im not really sure. I dont know why i did not work, did this change and things started working. I also added a karate test so we can discover if it happens again

PR #327, opened 2023-09-07
Np 5007 reorganize packages and classes

torarnet:


cristin-commons/src/main/java/no/unit/nva/cristin/common/client/CristinQueryApiClient.java
```diff
@@ -1,9 +1,10 @@
 package no.unit.nva.cristin.common.client;
 
+import no.unit.nva.client.GenericApiClient;
 import no.unit.nva.cristin.model.SearchResponse;
 import nva.commons.apigateway.exceptions.ApiGatewayException;
 
-public interface QueryApiClient<T, R> extends GenericApiClient {
+public interface CristinQueryApiClient<T, R> extends GenericApiClient {
```
torarnet, line 7: was more tightly coupled to cristin than i thought. gave it a name which reflected that

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/nva/ContributorRoleMapping.java
```diff
@@ -1,4 +1,4 @@
-package no.unit.nva.utils;
+package no.unit.nva.cristin.projects.model.nva;
```
torarnet, line 1: this was only used in project module. no need to have it in commons module

torarnet:
Mostly just movement between packages.

brinxmat (approved):
Good job on this; clear improvements and helpful notes!

PR #330, opened 2023-10-19
Np 45284 nr2 post patch person nvi data

torarnet:


anetteOlli (dismissed):

cristin-person/src/test/java/no/unit/nva/cristin/person/update/cristinPersonUpdate.feature
```diff
@@ -185,3 +186,77 @@ Feature: API tests for Cristin Person Update
...
+              'hello': 'world'
+          }
+      }
+    }
+    """
+    And request nviRequest
+    When method PATCH
+    Then status 400
+
+  Scenario: Update filters out nvi and returns status 400 indicating no supported fields when trying to verify nvi at another institution
```
torarnet, line 243: If clients wants to change any other fields it will be allowed, and nvi field will just be silently removed. but since nvi is filtered out here and it is the only field, it will complain that no supported fields are present

cristin-project/src/test/java/no/unit/nva/cristin/projects/create/createProject.feature
```diff
@@ -76,18 +75,3 @@ Feature: API tests for Cristin Project retrieve and search
     And match response.status == 400
     And print response.detail
 
-  Scenario: Create returns status 201 Created when input is valid and persisted in crisin
```
torarnet, line 79: We don't need all these creation of projects on each run, we already have a couple so those should be more than enough

torarnet:
Just to note: Karate tests can fail periodically without being an actual error. It depends on external systems and will not always go green. So it is still safe even if its red as long as one checks the action logs

cristin-person/src/test/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandlerTest.java
```diff
@@ -54,14 +64,51 @@
...
+        {
+            "preferredFirstName" : "Erik",
+            "nvi": {
+              "verifiedBy": {
+                "id": "https://api.dev.nva.aws.unit.no/cristin/person/12345",
+                "type": "Person",
+                "firstName": "Tor Inge",
+                "lastName": "Kristianslund"
+              },
+              "verifiedAt": {
```
StigNorland, line 90: Now, the naming futher up makes more sense... 
Can a person have roles from different organizations in the same context?
This is not something we can change, but  an entity (person) is partof X, X is partof Y, seems more logical and can be extended infinitely.

cristin-person/src/test/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandlerTest.java
```diff
@@ -190,8 +237,7 @@ void shouldReturnNoContentResponseWhenUserIsUpdatingTheirOwnPersonDataThatTheyAr
         throws IOException {
         var jsonObject = OBJECT_MAPPER.createObjectNode();
         jsonObject.put(ORCID, VALID_ORCID);
-        var gatewayResponse = queryWithPersonCristinIdButNoAccessRights(validPath,
-                                                                                          jsonObject.toString());
+        var gatewayResponse = queryWithPersonCristinIdButNoAccessRights(validPath, jsonObject.toString());
```
StigNorland, line 240: 👏

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/PersonSummary.java
```diff
@@ -8,7 +8,8 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.net.URI;
 
-public record PersonSummary(@JsonProperty(ID) URI id, @JsonProperty(FIRST_NAME) String firstName,
+public record PersonSummary(@JsonProperty(ID) URI id,
```
StigNorland, line 11: 👏

cristin-person/src/test/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandlerTest.java
```diff
@@ -241,7 +286,7 @@ void shouldAllowEmptyListOfEmployments() throws IOException {
         var jsonObject = OBJECT_MAPPER.createObjectNode();
         jsonObject.putArray(EMPLOYMENTS);
         var gatewayResponse =
-            sendQueryAsInstAdminWithSomeOrgId(jsonObject.toString(), getDummyOrgUri());
+            sendQueryAsInstAdminWithSomeOrgId(jsonObject.toString(), getDummyOrgUri185());
```
StigNorland, line 289: 👏

cristin-person/src/main/java/no/unit/nva/cristin/person/update/CristinPersonPatchJsonCreator.java
```diff
@@ -148,4 +161,48 @@ private void addBackgroundIfPresent() {
             output.set(BACKGROUND, input.get(BACKGROUND));
         }
     }
+
+    private void addNviIfPresent() {
+        if (input.has(NVI)) {
```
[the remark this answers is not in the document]
torarnet: has NVI = field is present in payload.
NVI is null = field is present in payload AND is null. (Very important as that means we should erase data. Field not present, don't do anything)
not has NVI = field NOT present at all

cristin-person/src/main/java/no/unit/nva/cristin/person/update/CristinPersonPatchJsonCreator.java
```diff
@@ -148,4 +161,48 @@ private void addBackgroundIfPresent() {
...
+            }
+
+            var nviNodeString = input.get(NVI).toString();
+
+            var nviInCristinFormat = attempt(() -> {
+                var parsedInput = OBJECT_MAPPER.readValue(nviNodeString, PersonNvi.class);
+                return OBJECT_MAPPER.readTree(nviToCristinFormat(parsedInput).toJsonString());
+            }).orElseThrow();
+
+            output.set(PERSON_NVI, nviInCristinFormat);
```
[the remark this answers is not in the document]
torarnet: set accepts a JsonNode, but we could stringify it. The only problem is I think ObjectNode handles JsonNode and String a little different. Maybe some " here and there that should or should not be there.

As long as this works as intended, is see no reason to change that, for now atleast

cristin-person/src/main/java/no/unit/nva/cristin/person/update/PersonPatchFieldFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    public PersonPatchFieldFilter(ObjectNode input) {
+        this.input = input;
+    }
+
+    /**
+     * Filters on given institution number. Payload needs to be validated first to verify it contains valid data.
+     * Automatically filters out data that cannot be updated without instNr if supplied instNr is null.
+     */
+    public PersonPatchFieldFilter filterOnInstNr(String instNr) {
+        if (input.has(NVI) && !input.get(NVI).isNull()) {
```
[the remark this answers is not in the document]
torarnet: see earlier comment

cristin-person/src/main/java/no/unit/nva/cristin/person/update/PersonPatchValidator.java
```diff
@@ -114,4 +127,32 @@ private static void parseKeywords(ObjectNode input) throws BadRequestException {
             throw new BadRequestException(COULD_NOT_PARSE_KEYWORD_FIELD);
         }
     }
+
+    private static void validateNviIfPresent(ObjectNode input) throws BadRequestException {
+        if (input.has(NVI) && !input.get(NVI).isNull()) {
```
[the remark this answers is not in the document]
torarnet: ditto

cristin-person/src/test/java/no/unit/nva/cristin/person/update/UpdateCristinPersonHandlerTest.java
torarnet: For now, verified by ANY institution is good enough. I have also included the top level organization in case someone decides to verify on sub level organization.

PR #340, opened 2023-12-06
Np 45763 retry on failed get

torarnet:


cristin-commons/src/main/java/no/unit/nva/cristin/common/client/ApiClient.java
```diff
@@ -135,19 +152,30 @@ public HttpResponse<String> fetchQueryResults(URI uri) throws ApiGatewayExceptio
...
 
         try {
-            return client.send(httpRequest, BodyHandlers.ofString(StandardCharsets.UTF_8));
-        } catch (HttpTimeoutException timeoutException) {
-            logError(ERROR_MESSAGE_BACKEND_FAILED_WITH_EXCEPTION, httpRequest.uri().toString(), timeoutException);
-            throw new GatewayTimeoutException();
-        } catch (IOException | InterruptedException otherException) {
-            logError(ERROR_MESSAGE_BACKEND_FAILED_WITH_EXCEPTION, httpRequest.uri().toString(), otherException);
-            throw new FailedHttpRequestException(ErrorMessages.ERROR_MESSAGE_BACKEND_FETCH_FAILED);
+            return fetchResponseWithRetry(httpRequest);
```
torarnet, line 158: the calling method `getSuccessfulResponseOrThrowException `is shared between GET, POST and PATCH. Should apply to most of the http requests across the whole proxy as ApiClient is extended everywhere

cristin-commons/src/main/java/no/unit/nva/cristin/common/client/ApiClient.java
```diff
@@ -206,10 +234,11 @@ private List<HttpResponse<String>> collectSuccessfulResponses(
         List<CompletableFuture<HttpResponse<String>>> responsesContainer) {
 
         return responsesContainer.stream()
-            .map(attempt(CompletableFuture::get))
-            .map(Try::get)
-            .filter(this::isSuccessfulRequest)
-            .collect(Collectors.toList());
+                   .map(attempt(CompletableFuture::get))
+                   .filter(Try::isSuccess)
```
torarnet, line 238: Fixing another issue when Try has an exception causing some requests to fail.

PR #348, opened 2024-01-09
Np 45619 person search without params

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/query/QueryCristinPersonHandler.java
```diff
@@ -111,6 +114,20 @@ protected void validateQueryParameterKeys(RequestInfo requestInfo) throws BadReq
         }
     }
 
+    @Override
+    protected String getValidName(RequestInfo requestInfo) throws BadRequestException {
+        var name = requestInfo.getQueryParameterOpt(NAME);
```
StigNorland, line 119: I have experimented with `Optional.ofNullable(value).stream()`

then you could have done this:
```
return Optional.ofNullable(name.).stream()
.filter(this::isValidQueryString)....
```

whether this is a good way of solving empty values, I don't know....

StigNorland (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/query/QueryCristinPersonHandler.java
torarnet: I wanted to return null if value is not supplied. if values is supplied, i validate and maybe throw exception. did not find a good way to do it by only declarative

anetteOlli (approved):

PR #354, opened 2024-02-14
np-46203: added support for D number when validating

torarnet:
bump dependency bekk commons


StigNorland (approved):

torarnet:
We calculated D number wrong. It seems to already be supported in both Cristin and Cristin Proxy. Closing PR

PR #356, opened 2024-02-26
Np 46206 list country codes

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/country/ListCountriesApiClient.java
```diff
@@ -0,0 +1,66 @@
...
+        super(httpClient);
+    }
+
+    public Countries executeRequest() throws ApiGatewayException {
+        Map<String, String> params = emptyMap();
+        var queryUri = createCristinQueryUri(params, CRISTIN_COUNTRIES_PATH);
+        var start = System.currentTimeMillis();
+        var response = queryUpstream(queryUri);
+        var countries = getCountries(response);
+        var totalProcessingTime = calculateProcessingTime(start, System.currentTimeMillis());
```
torarnet, line 42: im considering removing processingtime as I am caching the lambda response with cloudfront

brinxmat: Maybe only interesting for debug, so probably — we have other ways of getting this info now and as you point out the cache makes the info stale

cristin-person/src/main/java/no/unit/nva/cristin/person/country/ListCountriesHandler.java
```diff
@@ -0,0 +1,45 @@
...
+import nva.commons.apigateway.RequestInfo;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+import nva.commons.core.Environment;
+
+@SuppressWarnings("unused")
+public class ListCountriesHandler extends ApiGatewayHandler<Void, Countries> {
+
+    private final ListCountriesApiClient apiClient;
+
+    public ListCountriesHandler() {
```
brinxmat, line 19: We never return a single country value, only a list?

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinCountry.java
```diff
@@ -0,0 +1,19 @@
...
+
+import static no.unit.nva.cristin.model.JsonPropertyNames.CODE;
+import static no.unit.nva.cristin.model.JsonPropertyNames.CODE_ALPHA_3;
+import static no.unit.nva.cristin.model.JsonPropertyNames.NAME;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+import no.unit.nva.commons.json.JsonSerializable;
+import no.unit.nva.cristin.person.model.nva.Country;
+
+public record CristinCountry(@JsonProperty(CODE) String code,
```
brinxmat, line 11: In many ways, I'd prefer an id URI on each country that could be referenced and de-referenced

brinxmat: Possibly a new PR

cristin-person/src/main/java/no/unit/nva/cristin/person/country/ListCountriesHandler.java
torarnet: yes, we only list the whole dataset. it is only 255 results and i see no reason to do a lookup. it is also not updated often. if we want to have a lookup i have to implement in cristin api as well.

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinCountry.java
torarnet: do you have an example uri we should use?

cristin-person/src/main/java/no/unit/nva/cristin/person/country/ListCountriesApiClient.java
torarnet: i will remove this field

cristin-person/src/main/java/no/unit/nva/cristin/person/country/ListCountriesHandler.java
torarnet: if you want a lookup i can add a lookup in a new PR?

brinxmat (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinCountry.java
brinxmat: probably, the current /country/ + unique code (i.e. the ISO-code)

torarnet: iso 2 or 3 letter alpha3? we have both

PR #359, opened 2024-03-19
np-46448: use most recent org api version as default

torarnet:


torarnet:
I think I will wait with merge until tomorrow after notifying the team. One can always use earlier version by manually using Accept header

anetteOlli (approved):
Need to double check, but I think Scopus-import uses the new version already. Other than that, cristin and Brage migration does not use cristin-api.

PR #360, opened 2024-03-22
Np 46481 log if not getting access rights

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/query/organization/ListCristinOrganizationPersonsHandler.java
```diff
@@ -73,40 +77,20 @@ public ListCristinOrganizationPersonsHandler(CristinOrganizationPersonsClient ap
...
-        var numberOfResults = getValidNumberOfResults(requestInfo);
-        var name =
-            getNameIfPresent(requestInfo).orElse(null);
-        var sort =
-            getSortIfPresent(requestInfo).orElse(null);
-
-        var requestQueryParams =
-            buildParamMap(identifier, page, numberOfResults, name, sort);
-
-        return attempt(() -> getAuthorizedSearchResponse(requestInfo, requestQueryParams))
```
torarnet, line 92: If getAuthorizedSearchResponse throws something like ApiGatewayException, could that be swallowed by the attempt and then fallback to open data? Maybe that was an issue even if we have retries in our Http Client fetch?

anetteOlli (approved):

brinxmat (approved):

PR #365, opened 2024-04-05
Np 46506 change set to linkedset

torarnet:
To keep the sorting on org id as it is in Cristin we need to use datatype List instead of Set because Jackson will always treat a set as unsorted even if implementation is linkedhashset. Therefore i changed the datatype to keep the sort order when listing subunits. I overloaded the builder methods to keep backwards compatibility and avoid to much refactoring. We cannot sort on id ourselves in an easy way but the list should already be in sorted order when we get in in cristin. so we just need to keep this order when working with it.


torarnet:
Small PR but has some json testdata

torarnet:
Jackson always treated Set as unsorted even when implementing with LinkedHashSet. Needed to use List and used overloading for backwards compatibility

cristin-commons/src/main/java/no/unit/nva/model/Organization.java
```diff
@@ -167,11 +170,25 @@ public Builder withAcronym(String acronym) {
         }
 
         public Builder withPartOf(Set<Organization> partOf) {
+            if (nonNull(partOf)) {
+                this.partOf = new ArrayList<>(partOf);
+            }
+            return this;
+        }
+
+        public Builder withPartOf(List<Organization> partOf) {
```
brinxmat, line 179: You may be able to avoid this by using `withPartOf(Collection<Organization> partOf)` and the method above.

brinxmat (approved):

anetteOlli (approved):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/dto/v20230526/mapper/OrganizationFromUnitMapper.java
```diff
@@ -27,23 +29,23 @@ private Organization toOrganization(UnitDto unitDto) {
...
                    .withCountry(unitDto.getCountry())
                    .build();
     }
 
-    private List<Organization> unitsToOrganizations(List<UnitDto> unitDtos) {
+    private Set<Organization> unitsToSetOfOrganizations(List<UnitDto> unitDtos) {
         return unitDtos.stream()
                    .map(this::toOrganization)
-                   .collect(Collectors.toList());
+                   .collect(Collectors.toCollection(LinkedHashSet::new));
```
brinxmat, line 41: does `.distinct()` not fix the issue?

brinxmat (approved):

PR #367, opened 2024-04-10
Np 46489 enrich org search result

torarnet:
When doing organization query, add an optional param `fullTree` to decide if we want to enrich each search result with the tree of subunits and parent trail. Re-uses fetch api client as that has exactly what we want. We do a lookup using that on each search result to get more in depth data.

### DIFF
--- a/cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/v20230526/OrganizationEnricher.java
+++ b/cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/v20230526/OrganizationEnricher.java
@@ -17,9 +17,9 @@ import nva.commons.apigateway.exceptions.NotFoundException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class OrganizationSubUnitEnricher {
+public class OrganizationEnricher {
 
-    private static final Logger logger = LoggerFactory.getLogger(OrganizationSubUnitEnricher.class);
+    private static final Logger logger = LoggerFactory.getLogger(OrganizationEnricher.class);
 
     public static final String COULD_NOT_BE_FOUND_IN_UPSTREAM =
         "Organization from search result could not be found in upstream: {}";
@@ -29,16 +29,16 @@ public class OrganizationSubUnitEnricher {
     private final FetchApiClient<Map<String, String>, Organization> fetchClient;
     private final List<Organization> enrichedOrganizations;
 
-    public OrganizationSubUnitEnricher(List<Organization> inputOrganizations,
-                                       Map<String, String> queryParams,
-                                       FetchApiClient<Map<String, String>, Organization> fetchClient) {
+    public OrganizationEnricher(List<Organization> inputOrganizations,
+                                Map<String, String> queryParams,
+                                FetchApiClient<Map<String, String>, Organization> fetchClient) {
         this.inputOrganizations = inputOrganizations;
         this.queryParams = queryParams;
         this.fetchClient = fetchClient;
         enrichedOrganizations = new ArrayList<>();
     }
 
-    public OrganizationSubUnitEnricher enrich() throws ApiGatewayException {
+    public OrganizationEnricher enrich() throws ApiGatewayException {
         for (Organization organization : inputOrganizations) {
             var identifier = extractIdentifier(organization);
             var enrichParams = extractParams(identifier);
--- a/cristin-organization/src/test/java/no/unit/nva/cristin/organization/query/QueryCristinOrganizationHandlerTest.java
+++ b/cristin-organization/src/test/java/no/unit/nva/cristin/organization/query/QueryCristinOrganizationHandlerTest.java
@@ -23,6 +23,7 @@ import nva.commons.core.Environment;
 import nva.commons.core.attempt.Try;
 import nva.commons.core.ioutils.IoUtils;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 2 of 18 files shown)


torarnet:
Failing karate test is unrelated to this task

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/v20230526/OrganizationEnricher.java
```diff
@@ -0,0 +1,79 @@
...
+                logger.warn(COULD_NOT_BE_FOUND_IN_UPSTREAM, extractOrgId(organization));
+            }
+        }
+
+        return this;
+    }
+
+    public List<Organization> getResult() {
+        return enrichedOrganizations;
+    }
```
brinxmat, line 58: Since you only ever call `getResult()` after calling `enrich()`, maybe something like:

```

    public List<Organization> enrich() {
        return inputOrganizations.stream()
                .map(this::fetch)
                .filter(Optional::isPresent)
                .map(Optional::get)
                .collect(Collectors.toList());
    }

    private Optional<Organization> fetch(Organization organization) {
        var attempt = attempt(() -> fetchClient.executeFetch(extractParams(organization)));
        return attempt.isSuccess() ? Optional.of(attempt.get()) : handleException(attempt, organization);
    }

    private Optional<Organization> handleException(Try<Organization> attempt, Organization organization) {
        if (attempt.getException() instanceof NotFoundException) {
            logger.warn(COULD_NOT_BE_FOUND_IN_UPSTREAM, extractOrgId(organization));
            return Optional.empty();
        }
        throw new RuntimeException(attempt.getException());
    }
```

cristin-organization/src/test/java/no/unit/nva/cristin/organization/query/QueryCristinOrganizationHandlerTest.java
```diff
@@ -252,6 +261,55 @@ void shouldHaveCorrectCristinUriWithParamsOnVersionTwo() throws Exception {
...
+        var responseBody = gatewayResponse.getBodyObject(SearchResponse.class);
+        var actualHits = convertHitsToProperFormat(responseBody);
+
+        assertThat(gatewayResponse.getStatusCode(), equalTo(HTTP_OK));
+        assertThat(actualHits.get(0).getHasPart().size(), equalTo(8));
+        assertThat(actualHits.get(1).getHasPart().size(), equalTo(8));
+    }
+
+    @Test
+    void shouldDropEnrichmentWhenIdentifierNotFoundInUpstream() throws Exception {
```
brinxmat, line 290: Can you check that the message is logged?

cristin-organization/src/test/java/no/unit/nva/cristin/organization/query/QueryCristinOrganizationHandlerTest.java
```diff
@@ -252,6 +261,55 @@ void shouldHaveCorrectCristinUriWithParamsOnVersionTwo() throws Exception {
...
+        var input = handlerRequestWantingFullTree();
+        queryCristinOrganizationHandler.handleRequest(input, output, context);
+
+        var gatewayResponse = GatewayResponse.fromOutputStream(output, SearchResponse.class);
+        var responseBody = gatewayResponse.getBodyObject(SearchResponse.class);
+        var actualHits = convertHitsToProperFormat(responseBody);
+
+        assertThat(gatewayResponse.getStatusCode(), equalTo(HTTP_OK));
+        assertThat(actualHits.size(), equalTo(1));
+    }
```
brinxmat, line 311: Is there a test for e.g. gateway exception?

brinxmat (changes requested):

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/v20230526/OrganizationEnricher.java
torarnet: I did it this way in case i wanted to do more mutations on the state before retrieving the result, as is good practice in OOP. But if you think it is unnecessary, i can just return the object directly without having it in a separate method. If we wanted to add more logic later, having the result in a separate method could be a good thing...

torarnet: This will swallow exceptions related to upstream (for instance 502 Bad Gateway) and instead give a generic 500 Server Error even when there is no server error. That is why i had to do it without using functional style.

I can refactor if you think this is not a problem, will await your feecback

cristin-organization/src/test/java/no/unit/nva/cristin/organization/query/QueryCristinOrganizationHandlerTest.java
torarnet: Done

cristin-organization/src/test/java/no/unit/nva/cristin/organization/query/QueryCristinOrganizationHandlerTest.java
torarnet: Done

cristin-organization/src/main/java/no/unit/nva/cristin/organization/common/client/v20230526/OrganizationEnricher.java
brinxmat: Are you thinking of a case where you do e.g. 

```
            var organizationEnricher = new OrganizationEnricher(organizations, params, fetchClient);
            var organizations = organizationEnricher.enrich()
                 .translateLabels() // may be optionally and independently performed of "enrich()"
                 .addSubunitCount() // may be optionally and independently performed of "enrich()"
                 .getResult();
```

torarnet: Yes, something like that.

brinxmat (approved):

PR #368, opened 2024-04-10
fix: empty bucket before deleting stack #NP-46553

torbjokv:


.github/workflows/karate.yml
```diff
@@ -62,6 +62,14 @@ jobs:
         # Delete tests users after tests have run
       - name: Delete karate test users
         run: BASE_PATH=${{env.BASE_PATH}} ./gradlew deleteTestUsers --info
+      - name: Get NvaCristinIntermediateStorageBucket name
```
torarnet, line 65: dirty but works. I assume this is the best way to do it for now...

torarnet (approved):

PR #369, opened 2024-04-11
Np 46412 change country and keyword model

torarnet:
Changed datamodel for keyword and country to a model that is more in tune with our other data models for these types of data. Need to coordinate with frontend as this will break existing keyword query. The other endpoints, country list, keyword fetch and keyword creation are not used yet so should not be a problem


cristin-keyword/src/main/java/no/unit/nva/cristin/keyword/model/nva/Keyword.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @JsonCreator
+    public Keyword(@JsonProperty(CONTEXT) String context,
+                   @JsonProperty(TYPE) String type,
+                   @JsonProperty(ID) URI id,
+                   @JsonProperty(IDENTIFIER) String identifier,
+                   @JsonProperty(LABELS) Map<String, String> labels) {
+
+        super(context, type, id, identifier, labels);
+    }
```
brinxmat, line 30: Do we end up with a situation where we can create Keywords with a different type?

cristin-keyword/src/test/java/no/unit/nva/cristin/keyword/query/queryKeywords.feature
```diff
@@ -46,12 +46,14 @@ Feature: API tests for keywords query
     When method GET
     Then status 200
     And match response == '#object'
-    And match response['@context'] == '#present'
+    And match response['@context'] == 'https://bibsysdev.github.io/src/keyword-search-context.json'
```
brinxmat, line 49: Do you have such a context?

cristin-keyword/src/main/java/no/unit/nva/cristin/keyword/model/nva/Keyword.java
torarnet: No, not now at least

cristin-keyword/src/test/java/no/unit/nva/cristin/keyword/query/queryKeywords.feature
torarnet: No, can you help me with that one day?

anetteOlli (approved):

StigNorland (approved):

PR #373, opened 2024-04-30
Np 46395 more logging for authorized requests

torarnet:
Log id of created person, person employment and project resources. With this we can know what resource a client created by it's id in addition to what we already know, that being the cristin identifier and organization of the client.

### DIFF
--- a/cristin-commons/src/main/java/no/unit/nva/common/IdCreatedLogger.java
+++ b/cristin-commons/src/main/java/no/unit/nva/common/IdCreatedLogger.java
@@ -7,9 +7,9 @@ import no.unit.nva.model.UriId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class IdLogger {
+public class IdCreatedLogger {
 
-    private static final Logger logger = LoggerFactory.getLogger(IdLogger.class);
+    private static final Logger logger = LoggerFactory.getLogger(IdCreatedLogger.class);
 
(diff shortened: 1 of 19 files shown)


cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
StigNorland, line 1: Er ikke dette en god kandidat til en Record?

StigNorland (approved):
Fant ingenting 👍

cristin-person/src/main/java/no/unit/nva/cristin/person/model/nva/Person.java
torarnet: Det er nok mulig denne kunne vært en record ja. Jeg kunne testet det i en egen branch

cristin-commons/src/main/java/no/unit/nva/common/IdCreatedLogger.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class IdCreatedLogger {
+
+    private static final Logger logger = LoggerFactory.getLogger(IdCreatedLogger.class);
+
+    public static final String COULD_NOT_EXTRACT_IDENTIFIER_OF_NEWLY_CREATED_RESOURCE = "Could not extract identifier of newly created resource";
+    public static final String CLIENT_CREATED_RESOURCE_TEMPLATE = "Client created resource: %s";
+
+    public void logId(UriId uriId) {
+        attempt(() -> logCreatedIdentifier(uriId)).orElse(fail -> logCreatedError());
```
brinxmat, line 18: Is it possible to inject non-URI data here? If it is validated, I guess not

PR #379, opened 2024-06-06
external-style

sondrev:
Excplicit set ecplode: false for query parameters that do not support multiple listings (/users?id=3&id=4&id=5)

https://swagger.io/docs/specification/serialization/


cristin-commons/src/main/java/no/unit/nva/cristin/common/handler/CristinQueryHandler.java
```diff
@@ -39,6 +39,7 @@ protected void validateQueryParameterKeys(RequestInfo requestInfo) throws BadReq
 
     protected String getValidPage(RequestInfo requestInfo) throws BadRequestException {
         var page = requestInfo.getQueryParameterOpt(PAGE).orElse(FIRST_PAGE);
+        requestInfo.getMultiValueQueryParameter()
```
torarnet, line 42: what does this do?

sondrev: Oops. Looked up some code and committed by mistake. Removed

torarnet (approved):

torarnet:
Karate is failing because of unsupported characters in branch name

PR #380, opened 2024-06-14
Np 45735 treat nvi verified as regular verified

torarnet:


cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPerson.java
```diff
@@ -388,6 +391,21 @@ private Set<Employment> extractEmployments() {
...
+                              .map(PersonNvi::verifiedBy)
+                              .map(PersonSummary::id)
+                              .map(UriUtils::extractLastPathElement)
+                              .filter(Utils::isPositiveInteger)
+                              .isPresent();
+
+        if (nviVerified) {
+            return true;
+        } else {
+            return getIdentifiedCristinPerson();
```
monaullah, line 405: Would it be less "expensive" to check `getIdentifiedCristinPerson()` first?

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPerson.java
```diff
@@ -388,6 +391,21 @@ private Set<Employment> extractEmployments() {
                    .collect(Collectors.toSet());
     }
 
+    private Boolean extractVerified() {
+        var nviVerified = Optional.ofNullable(extractNvi())
```
monaullah, line 395: Not sure if I understand why this part is necessary:
```
.map(PersonNvi::verifiedBy)
.map(PersonSummary::id)
.map(UriUtils::extractLastPathElement)
.filter(Utils::isPositiveInteger)
```

monaullah: Would it be enough with `var nviVerified = Optional.ofNullable(extractNvi()).isPresent();`?
(I checked out your code and tried this, test was still green)

torarnet: I just wanted to make sure we actually have data and not just empty objects, but maybe im overthinking it

cristin-person/src/main/java/no/unit/nva/cristin/person/model/cristin/CristinPerson.java
torarnet: Yes, i will change the order

torarnet: Updated method

PR #382, opened 2024-06-18
Np 47035 refactor old code nr2

torarnet:
Just some small refactorings

### DIFF
--- a/cristin-commons/src/main/java/no/unit/nva/utils/AccessUtils.java
+++ b/cristin-commons/src/main/java/no/unit/nva/utils/AccessUtils.java
@@ -22,12 +22,15 @@ import static com.google.common.net.HttpHeaders.CONTENT_TYPE;
 import static java.net.http.HttpRequest.newBuilder;
 import static nva.commons.apigateway.AccessRight.MANAGE_CUSTOMERS;
 import static nva.commons.apigateway.AccessRight.MANAGE_OWN_AFFILIATION;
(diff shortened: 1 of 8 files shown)


cristin-commons/src/main/java/no/unit/nva/utils/AccessUtils.java
```diff
@@ -54,10 +50,7 @@ public class AccessUtils {
      * @throws ForbiddenException thrown when user is not authorized to access a user with IdentificationNumber
      */
     public static void validateIdentificationNumberAccess(RequestInfo requestInfo) throws ForbiddenException {
-        if (requesterHasNoAccessRightToUseNationalIdentificationNumber(requestInfo)) {
-            String nvaUsername = attempt(requestInfo::getUserName).orElse(fail -> null);
-            logger.warn(USER_DOES_NOT_HAVE_REQUIRED_ACCESS_RIGHT,
-                    nvaUsername, MANAGE_OWN_AFFILIATION);
+        if (!requesterIsUserAdministrator(requestInfo)) {
```
torarnet, line 53: if one checks this method `requesterIsUserAdministrator `one can see that it is equal to the one removed, apart from the negation which i have added here

torarnet: Kept the logging

brinxmat (approved):

PR #388, opened 2024-06-24
np-47035: refactor cristin person to record

torarnet:


brinxmat (approved):
Probably no reason to have any method with signature `methodName(List<Type> list)` when `methodName(Collection<Type> nameOfThings)` is equally valid. Sometimes we end up with set-to-list conversion before using methods simply because we have a non-permissive signature.

PR #400, opened 2024-08-01
Np 47255 update nva commons

torarnet:


cristin-commons/src/main/java/no/unit/nva/cristin/facet/CristinFacetUriParamAppender.java
```diff
@@ -12,7 +12,7 @@
 
 public class CristinFacetUriParamAppender {
 
-    public static final String PARAM_VALUE_DELIMITER = ",";
+    public static final String PARAM_VALUE_DELIMITER = ",|%2C";
```
StigNorland, line 15: PARAM_VALUE_DELIMITER is also defined other places in the codebase, concider to merge them.

cristin-commons/src/main/java/no/unit/nva/cristin/model/CristinQuery.java
```diff
@@ -96,11 +105,50 @@ public URI toNvaFacetURI() {
...
     public URI toURI() {
-        return
-            UriWrapper.fromUri(CRISTIN_API_URL)
-                .addChild(getCristinPath())
-                .addQueryParameters(toParameters())
-                .getUri();
+        var uri = UriWrapper.fromUri(CRISTIN_API_URL)
+                      .addChild(getCristinPath());
+
+        uri = appendParameters(toParameters(), uri);
```
StigNorland, line 111: due to bug in UriWrapper?

Should we rather fix UriWrapper?

cristin-commons/src/main/java/no/unit/nva/cristin/model/CristinQuery.java
```diff
@@ -96,11 +105,50 @@ public URI toNvaFacetURI() {
...
+            }
+        });
+
+        return multipleKeyMap;
+    }
+
+    private UriWrapper addQueryParameter(UriWrapper uri, NameValuePair pair) {
+        return UriWrapper.fromUri(uri.toString())
+                   .addQueryParameter(pair.getName(), pair.getValue());
+    }
```
truhacevkir, line 148: I have merged nva-commons, if you bump version to 1.40.7 I think you can remove `addQueryParameter` this method.

truhacevkir (approved):

cristin-person/src/main/java/no/unit/nva/cristin/person/employment/create/CreatePersonEmploymentHandler.java
StigNorland, line 1: My personal opinion is that all Handlers should be very sparse, with all operations being done by a service/client.
including all error handling, logging, validation and access right checks.
(testing handlers would then be to validate that they run) and all functionality would be tested one place, in the service/client)

cristin-commons/src/main/java/no/unit/nva/cristin/model/CristinQuery.java
torarnet: Will do, either in this PR or new one

cristin-person/src/main/java/no/unit/nva/cristin/person/employment/query/QueryPersonEmploymentHandler.java
```diff
@@ -44,7 +44,6 @@ public QueryPersonEmploymentHandler(QueryPersonEmploymentClient apiClient, Envir
     protected SearchResponse<Employment> processInput(Void input, RequestInfo requestInfo, Context context)
         throws ApiGatewayException {
 
-        AccessUtils.validateIdentificationNumberAccess(requestInfo);
         logger.info(LOG_IDENTIFIERS, extractCristinIdentifier(requestInfo), extractOrgIdentifier(requestInfo));
```
StigNorland, line 47: In another handler, you have moved logging to  validateRequest.
Done inside a common validate routine, it would be easier to maintain and handling would also be identical.

cristin-commons/src/main/java/no/unit/nva/cristin/model/CristinQuery.java
torarnet: Fixed in newest version that was merged today, as said in earlier comment. Will fix either in this or next PR

cristin-person/src/main/java/no/unit/nva/cristin/person/employment/create/CreatePersonEmploymentHandler.java
torarnet: I support the idea of having things like validation and access control int their own classes, and have already implemented that several places. But someone once said we should not leak handler objects like RequestInfo into other places than the handler, and we need that for things like query params.

The new method introduced that i had to implement was not something i had control over, i just had to implement it for contract reason.

cristin-person/src/main/java/no/unit/nva/cristin/person/employment/query/QueryPersonEmploymentHandler.java
torarnet: So move all common logic into one method/class?

StigNorland: It was more of an example of what can easily happen when code is implemented in the handler.
with regards to my earlier comments.

truhacevkir (approved):

monaullah (approved):

StigNorland (commented):
👍

PR #405, opened 2024-09-05
Np 47469 to strict query validation

torarnet:


torarnet:
> Does this change introduce the possibility that we can be sql-injected?

Yes, but it is either allowing that or keep a strict validation on query param which might cause some queries to fail.

Also, since this is just a proxy to cristin, cristin should handle it themselves.

Suggestions for improvements?

StigNorland (approved):

PR #406, opened 2024-09-17
np-47589: strip personal data from logs

torarnet:


ketilaa (approved):
This fixes the case(s) you found now. Can anything be done to avoid accidentally introducing similar "leaks" later?

torarnet:
> This fixes the case(s) you found now. Can anything be done to avoid accidentally introducing similar "leaks" later?

Only that we be vigilant and do the sanitation as close to the request as possible

PR #410, opened 2024-09-27
np-47689: Use field verified to know if we should keep id or not on p…

torarnet:
…roject contributor

we only want id on project contributors that are verified in cristin to make it easier for frontend

### DIFF
--- a/cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/adapter/CristinPersonToPerson.java
+++ b/cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/adapter/CristinPersonToPerson.java
@@ -20,16 +20,18 @@ public class CristinPersonToPerson implements Function<CristinPerson, Person> {
 
     private Person convert(CristinPerson cristinPerson) {
         return new Person(
-            convertId(cristinPerson.identifiedCristinPerson(), cristinPerson.cristinPersonId()),
+            extractIdIfIdentified(cristinPerson.identifiedCristinPerson(), cristinPerson.cristinPersonId()),
             cristinPerson.firstName(),
             cristinPerson.surname(),
             cristinPerson.email(),
             cristinPerson.phone());
     }
 
-    private URI convertId(Boolean identified, String cristinPersonId) {
-        return identifiedFieldNotPresent(identified) || identified
-                   ? getNvaApiId(cristinPersonId, PERSON_PATH_NVA) : null;
+    private URI extractIdIfIdentified(Boolean identified, String cristinPersonId) {
+        if (identifiedFieldNotPresent(identified) || identified) {
+            return getNvaApiId(cristinPersonId, PERSON_PATH_NVA);
+        }
+        return null;
     }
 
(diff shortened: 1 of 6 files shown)


cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/adapter/CristinPersonToPerson.java
```diff
@@ -19,15 +20,20 @@ public Person apply(CristinPerson cristinPerson) {
...
             cristinPerson.firstName(),
             cristinPerson.surname(),
             cristinPerson.email(),
             cristinPerson.phone());
     }
 
-    private URI convertId(String cristinPersonId) {
-        return getNvaApiId(cristinPersonId, PERSON_PATH_NVA);
+    private URI convertId(Boolean identified, String cristinPersonId) {
+        return identifiedFieldNotPresent(identified) || identified
```
brinxmat, line 31: So, if we don't have the data, or the person is identifier…we keep the URI?

brinxmat (dismissed):

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/adapter/CristinPersonToPerson.java
torarnet: Exactly. We dont want to remove the id if the field is missing. This makes it backwards compatible and if we want to enforce the field later it is a simple change in proxy

StigNorland (approved):
this is a no comments, comment

torbjokv (dismissed):

cristin-project/src/main/java/no/unit/nva/cristin/projects/model/cristin/adapter/CristinPersonToPerson.java
```diff
@@ -19,15 +20,20 @@ public Person apply(CristinPerson cristinPerson) {
...
+            convertId(cristinPerson.identifiedCristinPerson(), cristinPerson.cristinPersonId()),
             cristinPerson.firstName(),
             cristinPerson.surname(),
             cristinPerson.email(),
             cristinPerson.phone());
     }
 
-    private URI convertId(String cristinPersonId) {
-        return getNvaApiId(cristinPersonId, PERSON_PATH_NVA);
+    private URI convertId(Boolean identified, String cristinPersonId) {
```
LarsV123, line 30: Nitpick / non-blocking:
I needed some time and your comment below to understand what this does and why.
Can you think if a better structure and/or method name here, to make it more obvious?

The only suggestion I can think of is this, but I'm not sure it's actually better:
```
private URI getPersonUri(Boolean identified, String cristinPersonId) {
    if (isNull(identified) || identified) {
        return getNvaApiId(cristinPersonId, PERSON_PATH_NVA);
    }
    return null;
}
```

torarnet: Updated PR

brinxmat (approved):

PR #414, opened 2024-10-17
Revert "Noissue remove redundant depth param"

torarnet:
Reverts BIBSYSDEV/nva-cristin-service#412


torarnet:
Removed a query parameter which should not be removed as it actually was in use. Reverting the change

truhacevkir (approved):

ketilaa (approved):

PR #422, opened 2025-04-02
Migrate person response to valid json-ld

brinxmat:
- this is a breaking change for the API, but the data is unparseable and is therefore a bug (it will likely break loading into RDF models so no Person with an award can be expanded)
- `type` is a reserved word in NVA JSON-LD (it aliases the `@type` JSON-LD keyword)
- `type` was used to represent a relationship between an award and what the award was awarded for; `type`-> `awardFor`
- JSON-LD context did not exist, now it does; `https://example.org/person-context.json`-> `https://bibsysdev.github.io/src/person-context.json`
- New example: https://tinyurl.com/preview/27c39dsn
- Fixed one checkstyle warning


ketilaa (dismissed):
Go for it!

ketilaa (approved):

PR #451, opened 2026-08-11
fix(cristin-person): return 307 for person merged into another #NP-51564

torbjokv:
Cristin answers `307` with a relative location when a person has been merged into another one. The JDK client follows redirects, so `GET /cristin/person/211415` returned person `5647` as `200 OK` under the wrong URI. The person endpoint now redirects to the canonical person instead.

```
GET /cristin/person/211415   → 307, Location: .../cristin/person/5647
GET /cristin/person/0359084  → 307, Location: .../cristin/person/359084
GET /cristin/person/{orcid}  → 200 + person data
```

- Compare the requested identifier against `cristin_person_id` in the response body, which is Cristin's own answer to which person this is. That covers both a merged identifier and a non-canonical spelling of the same one, for the authorized and the unauthorized fetch alike
- Add `TemporaryRedirectException`, a concrete `RedirectException`, which `ApiGatewayHandler` turns into a `307` with a `Location` header pointing at the NVA person URI
- Answer `200` for ORCID lookups. Cristin implements those as a `307` as well (`/v2/persons/ORCID:0000-0003-0013-2874 → 307, location: 5647`), but an ORCID is a different kind of key rather than a non-canonical spelling of a Cristin identifier
- Log the redirect without the stack trace `RestRequestHandler` would otherwise emit as a `WARN` on every merged-person lookup
- Document the `307` and its `Location` header on `GetPersonById` in the OpenAPI spec

**Before merging:** `scopus-import` in `nva-publication-api` builds its `HttpClient` without `followRedirects` and turns any non-200 into `Optional.empty()` (`CristinConnection:52,133`), so a merged person silently loses its Cristin enrichment. The ORCID fallback in `CristinPersonRetriever` covers contributors that have an ORCID, but not the rest. That client needs `followRedirects(Redirect.ALWAYS)` deployed first.

Query and search are untouched, so enrichment there still follows redirects silently. Cristin's search does not normally return merged persons.

https://sikt.atlassian.net/browse/NP-51564

### DIFF
--- a/cristin-person/src/main/java/no/unit/nva/cristin/person/client/CristinPersonApiClient.java
+++ b/cristin-person/src/main/java/no/unit/nva/cristin/person/client/CristinPersonApiClient.java
@@ -297,38 +297,10 @@ public class CristinPersonApiClient extends ApiClient
     return toCristinPerson(identifier, fetchGetResult(uri));
   }
 
-  private CristinPerson toCristinPerson(String identifier, HttpResponse<String> response)
-      throws ApiGatewayException {
-    checkHttpStatusCode(getNvaApiId(identifier, PERSON), response.statusCode(), response.body());
-    throwRedirectWhenPersonIsMergedIntoAnother(identifier, response);
-    return getDeserializedResponse(response, CristinPerson.class);
-  }
-
   private URI getCorrectUriForIdentifier(String identifier) {
     return isOrcid(identifier)
         ? CristinPersonQuery.fromOrcid(identifier)
         : CristinPersonQuery.fromId(identifier);
-  }
-
-  private void throwRedirectWhenPersonIsMergedIntoAnother(
-      String requestedIdentifier, HttpResponse<String> response) throws TemporaryRedirectException {
-
-    var redirectedToIdentifier = extractPersonIdentifier(response.uri());
-    if (upstreamRedirectedToAnotherCristinPerson(requestedIdentifier, redirectedToIdentifier)) {
-      logger.info(LOG_PERSON_MERGED_INTO_ANOTHER, requestedIdentifier, redirectedToIdentifier);
-      throw new TemporaryRedirectException(getNvaApiId(redirectedToIdentifier, PERSON));
-    }
-  }
-
-  private boolean upstreamRedirectedToAnotherCristinPerson(
-      String requestedIdentifier, String redirectedToIdentifier) {
-    return isPositiveInteger(requestedIdentifier)
-        && isPositiveInteger(redirectedToIdentifier)
-        && Integer.parseInt(requestedIdentifier) != Integer.parseInt(redirectedToIdentifier);
-  }
-
-  private String extractPersonIdentifier(URI uri) {
-    return nonNull(uri) && isNotBlank(uri.getPath()) ? extractLastPathElement(uri) : null;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


docs/cristin-proxy-swagger.yaml
```diff
@@ -1413,6 +1413,17 @@ paths:
...
+          description: >-
+            Temporary Redirect. The requested person has been merged into another person and
+            the Location header points to the person to use instead
+          headers:
+            Location:
+              description: Uri of the person the requested person has been merged into
+              schema:
+                type: string
+                format: uri
+              example: "https://api.nva.unit.no/cristin/person/5647"
```
LarsV123, line 1426: Does this break `scopus-import` in `publication-api`? I don't think that client follows redirects, it just treats non-200 as a failure and silently ignores it.

cristin-person/src/main/java/no/unit/nva/cristin/person/client/CristinPersonApiClient.java
```diff
@@ -347,6 +357,40 @@ private URI extractFirstUriFromListOfCristinPersons(List<CristinPerson> cristinP
...
+
+  private String extractPersonIdentifier(URI uri) {
+    return isCristinPersonResource(uri) ? extractLastPathElement(uri) : null;
+  }
+
+  private boolean isCristinPersonResource(URI uri) {
+    return nonNull(uri)
+        && nonNull(uri.getPath())
+        && CRISTIN_PERSON_PATH.matcher(uri.getPath()).matches();
+  }
```
LarsV123: Does this handle differences in casing or trailing slashes etc? Should we normalize the URLs before doing the check?

cristin-person/src/main/java/no/unit/nva/cristin/person/fetch/FetchCristinPersonHandler.java
```diff
@@ -72,6 +74,17 @@ protected Integer getSuccessStatusCode(Void input, Person output) {
     return HttpURLConnection.HTTP_OK;
   }
 
+  @Override
+  protected void handleExpectedException(Context context, Void input, ApiGatewayException exception)
+      throws IOException {
+
+    if (exception instanceof RedirectException) {
+      writeExpectedFailure(input, exception, context.getAwsRequestId());
```
LarsV123, line 83: This should probably have equivalent logging to the method it overrides, so that a failure isn't silently swallowed.

cristin-person/src/main/java/no/unit/nva/cristin/person/client/CristinPersonApiClient.java
```diff
@@ -347,6 +357,40 @@ private URI extractFirstUriFromListOfCristinPersons(List<CristinPerson> cristinP
...
+      logger.info(LOG_PERSON_MERGED_INTO_ANOTHER, requestedIdentifier, redirectedToIdentifier);
+      throw new TemporaryRedirectException(getNvaApiId(redirectedToIdentifier, PERSON));
+    }
+  }
+
+  private boolean upstreamRedirectedToAnotherCristinPerson(
+      String requestedIdentifier, String redirectedToIdentifier) {
+    return isPositiveInteger(requestedIdentifier)
+        && isPositiveInteger(redirectedToIdentifier)
+        && Integer.parseInt(requestedIdentifier) != Integer.parseInt(redirectedToIdentifier);
```
LarsV123: Can Cristin identifiers ever be zero-padded and would that break something here? This treats `/persons/0123` as the same as `/persons/123`.

LarsV123 (changes requested):

docs/cristin-proxy-swagger.yaml
torbjokv: working on a PR to fix this. https://github.com/BIBSYSDEV/nva-publication-api/pull/2561

cristin-person/src/main/java/no/unit/nva/cristin/person/client/CristinPersonApiClient.java
torbjokv: It didn't. Now comparing the requested identifier against cristin_person_id from the response body instead of the URI, so there's no path left to normalize.

cristin-person/src/main/java/no/unit/nva/cristin/person/fetch/FetchCristinPersonHandler.java
torbjokv: Agreed, added logger.info with the redirect target.

cristin-person/src/main/java/no/unit/nva/cristin/person/client/CristinPersonApiClient.java
torbjokv: We changed from implicit to explicit canonicalisation: the canonical id was always in the response body, now /person/0359084 returns 307 to /person/359084. Done by comparing the requested identifier against cristin_person_id from the body (CristinPersonApiClient:373) instead of the URI.

LarsV123 (approved):

